PR #1, opened 2026-09-18
feat: extract search, jobs, events and mail packages

ManavA:
## Summary

Extracts four reusable packages from GoReal's generic infrastructure, per ARCHITECTURE.md's layering and "usable alone" rule:

- **search** — a generic document index behind an `Index` interface. Two implementations: `PostgresIndex` (single JSONB table + tsvector, no external service, suitable for small tables) as the default, and `Searcher` (Meilisearch: facets, typo tolerance, synonyms, settings-drift detection) as the opt-in upgrade. Both share the same `Query`/`Filter`/`SortField` builder.
- **jobs** — `Outcome`/`Complete` compute a run's status word from its own counts instead of a hardcoded literal, so a run that had work and completed none of it cannot report success. `Scheduler` runs entries in-process on their own tickers — no external scheduler service required. `Guard`/`Idempotent` skip already-completed work. `Trigger`/`CloudRunTrigger` are an optional way for one job to start another directly on Cloud Run. `WatchdogStatus`/`CompleteWatchdog` give a job that verifies another job's effect the same three-state result as `Outcome` (verified / diverged / inconclusive).
- **events** — `Publisher`/`Subscriber` interfaces. `InMemoryBus` is the default, no broker required. `PubSubPublisher`/`PubSubSubscriber` (Pub/Sub v2) are the opt-in cross-process transport.
- **mail** — `Sender` interface. `LogSender` (slog, no external service) is the default; `PostmarkSender` is the opt-in provider-backed implementation, including the `res.ErrorCode` check the underlying client library omits and a startup check (`CheckTemplates`) that confirms required template aliases exist in the account. `mail/deliverycheck` reconciles a job's delivery claims against a provider's own records. `mail/testing` provides a `Recorder` for tests.

## Test counts

119 passing tests across all six packages (`jobs`, `events`, `mail`, `mail/deliverycheck`, `mail/testing`, `search`), plus two `live`-tagged Postgres integration tests in `search` gated on `KEEL_TEST_POSTGRES_DSN` (skip loudly when unset). `go build`, `go vet`, `golangci-lint run ./...` (0 issues), and `govulncheck ./...` (0 reachable vulnerabilities) all pass. `gofmt -l .` is clean.

## Mutation check

Mutated a load-bearing line in each package, confirmed the test suite went red, then reverted:

| Package | Mutation | Result |
|---|---|---|
| jobs | `Outcome.DidNothing`: `Succeeded <= 0` → `Succeeded < 0` | red: `TestComplete_MessageChoiceTracksOK` |
| events | `InMemoryBus.Subscribe`: skip calling `handler` on each message | red: 3 tests (delivery, fan-out, error-doesn't-stop-loop) |
| mail | `classifyPostmarkError`: `code == postmarkInactiveRecipient` → `!=` | red: 2 tests (inactive-recipient, other-codes-not-permanent) |
| mail | `CheckTemplates`: `templateExists` always returns `true` | red: `TestCheckTemplates_NamesEveryMissingAlias` |
| mail/deliverycheck | `Reconcile`: disabled the `ControlPassed` guard | red: `TestAFailedNegativeControlVoidsTheWholeRun` |
| search | `numericFieldClause`: always emit `=` regardless of the operator | red: `TestBuildPostgresWhere_RangeFiltersAreCastToNumeric` |

## Deliberately left out, and why

- **No listing/MLS/domain types anywhere.** `search.Document` is `map[string]any` or a one-method interface; nothing here knows what a caller is indexing.
- **No Fair Housing or other content-policy filtering** in `mail` — that's `textpolicy/` (extractor B); a transport package shouldn't decide what's sayable.
- **No sender-persona rotation, brand copy, or template models** in `mail` — those are product decisions for the service sending the mail, not the library delivering it.
- **No geo search** in `search` — Meilisearch supports it, but building a portable `Filter`/`SortField` shape for `_geoBoundingBox`/`_geoPoint` was out of scope for this pass; `OpRaw` is the escape hatch for a caller who needs it against the Meilisearch backend specifically.
- **No in-memory Meilisearch fake** — none exists, so `Searcher`'s tests fake the `clientAPI`/`indexAPI` interfaces to exercise the query builder and setup/drift logic; anything needing a live server would be a `live`-tagged test (none needed yet, since the fakes cover the logic).
- **Cloud Run job entrypoint conventions** (the AST-based "no hardcoded status literal" scanner, the job-name/monitor-expectation consistency check) were left in the source service rather than ported here — they read `cmd/*/main.go` paths and a domain-specific job registry that don't generalize.

## Dependency notes

`go.mod` picked up `meilisearch-go`, `keighl/postmark`, `cloud.google.com/go/pubsub/v2`, `jackc/pgx/v5`, `golang.org/x/oauth2`, and `testify`. `cloud.google.com/go/pubsub` (v1, deprecated) was avoided in favor of v2 to keep `govulncheck`/staticcheck clean. `grpc` and `golang.org/x/text` were bumped past their transitively-required minimums to clear two reachable vulnerabilities flagged by `govulncheck` in the pubsub/v2 dependency chain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011kwifre98AiLQDrdqZX6no


ManavA:
## Independent review — `b19b536`

**Verdict: CHANGES REQUESTED.** Three blockers, all verified by running code rather than reading it. Everything else below is a should-fix or a note.

The extraction itself is solid: layering is exactly right, the Postgres injection story is genuinely safe (and I tried hard to break it), and all six claimed mutations reproduce. The blockers are concentrated in default paths — a panicking job, the default mail sender, and a doc that overclaims a quoting guarantee.

### What I verified green

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test ./... -race` | clean, 121 test funcs (119 non-live + 2 live) |
| `golangci-lint run ./...` @ **v2.6.0** (Makefile/CI pin) | **0 issues** |
| `govulncheck ./...` | 0 reachable |
| `gofmt -l .` | clean |
| `go mod tidy` | no-op; `oauth2` correctly direct (`jobs/trigger.go:12`) |
| `gitleaks detect` — working tree **and** all 3 commits of history | no leaks |
| Import levels (`go list -deps`) | **zero** keel-internal imports in all 6 packages |
| Each package alone in a scratch module | all 6 compile |
| Live Postgres tests vs real PG 16 | both pass; skip loudly naming `KEEL_TEST_POSTGRES_DSN` |

> Note: `golangci-lint@latest` reports 30 `goconst` issues in test files. At the **pinned v2.6.0** it is 0, so this PR is clean against the gate as configured — flagging only because the pin will drift.

### Mutation table

Your six, re-applied from scratch, then reverted:

| # | Package | Mutation | Result |
|---|---|---|---|
| M1 | jobs | `DidNothing`: `Succeeded <= 0` → `< 0` | **killed** — 3 tests |
| M2 | events | `Subscribe` never calls `handler` | **killed** — 3 tests |
| M3 | mail | `classifyPostmarkError`: `==` → `!=` | **killed** — 2 tests |
| M4 | mail | `templateExists` always `true` | **killed** — `TestCheckTemplates_NamesEveryMissingAlias` |
| M5 | deliverycheck | `ControlPassed` guard disabled | **killed** — `TestAFailedNegativeControlVoidsTheWholeRun` |
| M6 | search | `numericFieldClause` always emits `=` | **killed** — `TestBuildPostgresWhere_RangeFiltersAreCastToNumeric` |

Mine:

| # | Mutation | Result |
|---|---|---|
| N1 | `PostgresIndex` silently ignores `OpRaw` instead of refusing | **killed** — `TestBuildPostgresWhere_RawFilterIsAnError` |
| N2 | Drop the default `ORDER BY id` (search/postgres.go:270) | **SURVIVED** |
| N3 | tsvector query config `simple` → `english` | killed |
| N4 | `quoteIdent` stops doubling embedded quotes | killed |
| N5 | `Idempotent` treats a `Done` error as not-done | killed |
| N6 | `FormatAddress` stops treating `,` as special | killed |
| N8 | `PruneStale` always reports 0 deleted | **SURVIVED** (default suite; caught only under `-tags=live`) |

---

## Blockers

### B1 — `jobs`: a panicking job takes down the process, and the whole Scheduler with it

`Runner.Run` (jobs/runner.go:39) calls `fn(ctx)` with no `defer recover()`, and `Scheduler.runLoop` (jobs/scheduler.go:85) adds none. Verified — this is not theoretical:

```
### TestProbe_SchedulerSurvivesPanickingEntry
panic: entry exploded
FAIL	adversarial	0.007s
```

The test binary died. Every sibling entry registered on that `Scheduler` died with it. For a package whose stated selling point is "runs registered `Entry` values on their own tickers inside the calling process. No external scheduler service is required" (jobs/doc.go:25-29), one bad job killing every other job in the shared process is the defining failure mode, and it is the one thing an external scheduler would have survived.

Suggested: `defer recover()` in `Runner.Run`, record `Outcome{Fatal: true}`, log the value and stack, return the non-zero exit code. `runLoop` then keeps ticking.

### B2 — `mail`: `LogSender` writes the recipient and the entire template model to the log at INFO

mail/logsender.go:34 logs `"to", to` and `"model", templateModel`. Verified output:

```json
{"level":"INFO","msg":"mail not sent: no provider configured",
 "to":"victim@example.com","template":"password-reset",
 "model":{"action_url":"https://app.example.com/reset?t=SECRET-TOKEN-abc123",
          "name":"Jane Doe","reset_token":"SECRET-TOKEN-abc123"}}
```

`LogSender` is documented as the default "when no mail provider is configured" (mail/doc.go:24-26) — so it is exactly what runs in production on a misconfiguration, writing password-reset tokens, magic links and recipient addresses into whatever log sink the service ships to. The review brief allows "say what it logs and why that is acceptable for a dev sender"; the package doc currently says neither.

Suggested: log the alias plus a recipient domain or digest, drop the model by default, and put the full model behind an explicit opt-in (`LogSenderOptions.LogTemplateModel bool`).

### B3 — `search`: Meilisearch range ops do not quote their value, and the package doc says they do

search/doc.go:50-51 states: "*[Filter] constructors and [buildMeiliFilters] quote every string value*". search/filters.go:39-45 interpolates `%v` bare for `OpGte`/`OpLte`/`OpGt`/`OpLt`, while `OpEq`/`OpNeq`/`OpIn` go through `QuoteFilterValue`. Verified against a stub Meilisearch:

```
Gte("price", `0 OR secret = 1`)  ->  ["price >= 0 OR secret = 1"]     <- injected verbatim
Eq ("price", `0 OR secret = 1`)  ->  ["price = \"0 OR secret = 1\""]   <- correctly quoted
```

`Filter.Value` is `any`, and range filters are the ones most often fed straight from caller-facing query-string input (`min_price=…`), so a string reaching a range op is reachable rather than hypothetical. It corrupts the intended clause rather than escaping the AND chain, but the doc asserts a safety property the code does not provide.

Suggested: route range values through `meiliValue` too, or reject a non-numeric value for a range op with a named error.

---

## Should fix

**S1 — `mail`: `CheckTemplates` reports a bad token or a provider outage as "template not found".** mail/templates.go:92-94 returns `(false, nil)` for every non-2xx. Verified — 401, 422, 500 and 503 all produce the identical:

```
mail: 1 required template(s) not found in this Postmark account: welcome
```

This is the same "could not measure" / "measured, and absent" conflation that `jobs.Outcome` and `deliverycheck` exist to prevent — in the same PR. A startup check that says "your template is missing" when the real problem is an invalid server token sends an operator to the wrong place. Suggested: distinguish 401/403 and 5xx as an error to propagate, and keep "not found" for the 404/422 shape.

**S2 — `events`: the two implementations disagree on `[]byte` payloads.** `marshalEvent` (events/subscriber.go:161) passes `[]byte` through raw; `PubSubPublisher.Publish` and `NoopPublisher.Publish` use plain `json.Marshal`. Verified:

```
InMemoryBus delivered:  {"hello":"world"}
PubSub/Noop recorded:   "eyJoZWxsbyI6IndvcmxkIn0="
```

events/doc.go:24-26 promises a caller "can switch between InMemoryBus and Pub/Sub without changing its own code."

**S3 — `events`: delivery semantics are not documented, and drops are silent.** `InMemoryBus` is at-most-once — a full 16-slot buffer drops on the floor with no log and no counter (events/subscriber.go:111-116). Pub/Sub is at-least-once. My probe: of 100 messages published past a blocked handler, **17 arrived and 83 vanished silently**. Publish's own doc admits the drop, but `events/doc.go` — where someone choosing an implementation actually looks — says only that they are interchangeable. A handler must be idempotent for one and will silently lose data on the other. At minimum say so in the package doc; ideally count or log the drop.

**S4 — `search`: `Searcher` ignores `context.Context` everywhere.** `Search`, `Health`, `SetupIndex` and `DocumentCount` all take `_ context.Context`. ARCHITECTURE.md is explicit: "a function that takes a `context.Context` and ignores it is worse than one that does not take it at all." `PostgresIndex` honours it (verified: a cancelled context errors), so the two `Index` implementations differ in cancellation semantics — a caller that switches backends loses request cancellation and timeouts silently. I accept this is forced by meilisearch-go v0.26.1 (only `awaitTask` gets a real context, via `WaitParams`), but it is undocumented. Say so on `Searcher`.

**S5 — "optional external service" holds at the API level but not the dependency level.** Transitive package counts:

| Package | deps | pulled in for |
|---|---|---|
| `events` | **299** | Pub/Sub v2 + gRPC + OTel + protobuf — even for an `InMemoryBus`-only caller |
| `search` | 62 | Meilisearch + fasthttp — even for a `PostgresIndex`-only caller |
| `jobs` | 24 | `oauth2/google` — even for an `Outcome`/`Complete`-only caller |

ARCHITECTURE.md's own test is "`go get` it into a service that has no HTTP server at all and it compiles". It compiles — at 299 transitive packages for the in-process default. The repo already demonstrates the fix (`mail/deliverycheck`, `mail/testing` as subpackages): move `PubSubPublisher`/`PubSubSubscriber` to `events/pubsub`, `CloudRunTrigger` to `jobs/cloudrun`, `Searcher` to `search/meili`.

**S6 — `jobs`: `Idempotent` keys are one flat global namespace.** Verified: two unrelated jobs both using the natural key `"2026-09-17"` — the second never runs and reports success. Nothing on `Guard` or `Idempotent` warns about this, and a date or a batch id is the obvious key to reach for. Either namespace by job name in the API or document the requirement that keys be globally unique.

**S7 — `search`: doc.go:11 calls the Postgres free-text search "ILIKE-style".** It is `to_tsvector('simple')` + `plainto_tsquery('simple')`. Verified against real Postgres:

| query | hits vs document "Charming bungalow" |
|---|---|
| `bungalow` | 1 |
| `bung` (substring) | **0** — ILIKE would match |
| `BUNGALOW` | 1 |
| `charm` (stem of "Charming") | **0** — `simple` does not stem |

Whole-lexeme, unstemmed matching is a reasonable choice, but "ILIKE-style" describes something else. Also worth noting the `'simple'` config is hard-coded in both places with no `PostgresConfig` knob, so there is no way to opt into stemming or stop-words.

**S8 — `search`: `Query.Offset` is not clamped the way `Limit` is.** `limit <= 0` gets a default (search/postgres.go:273-276); a negative `Offset` goes straight to Postgres and surfaces as `search: ERROR: OFFSET must not be negative (SQLSTATE 2201X)`.

**S9 — `search`: a numeric filter over a mixed-type JSONB field returns a raw driver error that echoes the data.** `Gte("price", 50)` where one document has `"price": "not a number"` yields `count matching documents: ERROR: invalid input syntax for type numeric: "not a number" (SQLSTATE 22P02)`. A whole-query failure caused by one bad row, with the offending value in the message.

**S10 — two survived mutations.** `N2`: deleting the default `ORDER BY id` (search/postgres.go:270) leaves the suite green, although its own comment correctly explains it is load-bearing for paging correctness. `N8`: `PruneStale` returning 0 instead of `len(stale)` is green in the default suite and caught only under `-tags=live`.

**S11 — minor.** `CheckTemplates` does not URL-escape the alias — `"weird alias/with slash"` produces the request path `/templates/weird alias/with slash`. And `mail/testing` is package `testing`, so any caller must alias it next to stdlib `testing`.

---

## Writing style

Better than the "some inline comments may still be informal" concession implied — the informal ones are mostly in *doc* comments, not inline. Two clear violations of the simple/direct/professional rule:

- **`mail/deliverycheck/reconcile_test.go:152-155`** — "*a check that **cries wolf** gets switched off, and then it is exactly as useful as no check at all.*" A fable allusion plus an aphorism about checks in general. Suggested: "a check that reports false divergences will be disabled, and a disabled check verifies nothing."
- **`search/settings_drift.go:138-139`** — "*collapsing them turns a failure to measure into a **clean bill of health**.*" Stock idiom. Suggested: "…into a reported pass."

Borderline, listed so you can decide once rather than per-line — two repeated habits:

1. *Personifying the code.* "the exit code the run **earned**" appears 5×: `jobs/outcome.go:75`, `jobs/outcome.go:86`, `jobs/runner.go:37`, `jobs/watchdog.go:44`, `jobs/doc.go:19-20`. One find-and-replace to "the exit code for this run" if you want it gone. Same family: `jobs/outcome.go:46-47` ("a hardcoded \"success\" cannot **see**"), `jobs/watchdog.go:34-35` ("the expected, **honest** shape"), `events/subscriber.go:100-102` ("rather than **pretending** unbounded delivery is free").
2. *A rhetorical aside after an em-dash, closing an otherwise plain sentence.* `mail/deliverycheck/reconcile.go:43-44` ("— which is **NOT good news**") is the most conversational line in the non-test code; also `search/settings_drift.go:99-100` ("— loud, but for the wrong reason"), `jobs/outcome.go:29-30` ("the **interesting** cases"), `search/searcher.go:271-275` ("the **cheapest possible** check … stale ones **forever**"), `search/meili_search.go:62-63` ("the **vast majority** of queries actually want").

Two smaller ones: `mail/postmark.go:85` is the only first-person voice in the non-test code ("a mistake in **our own** configuration"); "dead pipeline … healthy" is a medical metaphor used twice (`mail/deliverycheck/reconcile.go:70-72` and `reconcile_test.go:60-62`), though it is common enough ops vocabulary to defend.

Explicitly **not** flagged, because they are hypothetical failure-mode exposition rather than anecdotes: `jobs/doc.go:7-16` (the literal-`"success"` walkthrough) and `mail/doc.go:7-15` (the keighl/postmark 1101 explanation). Those are the two best-written passages here. Mechanical sweep over all 48 files found no second-person lecturing, no "obviously/simply/clever/elegant/magic", no exclamation marks, no rhetorical questions.

## Domain types

**Clean on the blocking axis.** Every declared type, struct field, constant, function and interface across the four packages is generic infrastructure — `Document`, `MapDocument`, `Index`, `Query`, `Filter`, `SortField`, `Outcome`, `Entry`, `Guard`, `Publisher`, `Subscriber`, `Sender`, `Claim`, `Verdict`, `Recorder` and so on. The `PostgresIndex` schema is `id` / `document` / `search_text`. Two greps that look like hits are not: `broker` in `events/subscriber.go` is the *message* broker, and `idx` in `search/postgres_query.go:98-101` is a local for "index", not the IDX feed.

Remaining real-estate flavour is confined to **test fixtures and doc examples**, almost entirely in `search/` (`"Charming bungalow"`, `"Alameda"`, `"Redwood City"`, `"sf" -> "san francisco"`, `price`/`sqft`/`beds`). Non-blocking under the stated rule, but two would catch a reader's eye and read as leakage:

- `search/postgres_query_test.go:125` — `quoteIdent("listings")`
- `search/postgres_live_test.go:48-50` — the bungalow/condo/Alameda document set, which together read as a property database

Swapping `"listings"` for `"documents"` and neutralising those three fixtures removes nearly all the visible domain flavour for very little churn. `jobs/` and `events/` are already fully neutral (`"orders"`, `"widget"`, `"reindex"`).

## Verified good, for the record

- **Postgres SQL injection: genuinely safe, and I tried.** Both filter **values** and filter **field names** are bound parameters (`document->>$1`), including in `ORDER BY`. Against real Postgres 16 I sent `city'; DROP TABLE …; --` and `' OR 1=1 --` as values, as field names, and as sort fields: every one inert, zero rows matched, table intact. The `->>`-takes-a-parameter comment (search/postgres_query.go:14-18) is correct and it is a nice trick — worth keeping the comment, since the next reader will assume it needs an identifier whitelist.
- `OpRaw` reaching `PostgresIndex` is refused with a message that names the field and says what to do instead — and N1 proves a test holds that line.
- `EnsureSchema` is idempotent: three re-runs, data intact.
- The `Scheduler` **skips** rather than double-runs an overlapping tick (a job 5× slower than its interval: max concurrency 1, 7 runs in 220ms). Correct behaviour — please document it, since it is the first question an operator asks.
- `deliverycheck`'s negative control is real and load-bearing, and the empty-claims and settle-window paths both return Inconclusive rather than OK.
- `PostmarkSender` 200-with-`ErrorCode` is genuinely covered — the stub writes an explicit `http.StatusOK`.
- The deliberately-left-out list matches the code: no `_geo` anywhere, no entrypoint scanners, no brand copy.
- The dependency-bump note checks out directionally: `grpc v1.83.1` and `x/text v0.39.0` are pinned by keel itself, above the graph's own maxima (`v1.79.3` / `v0.37.0`); `otel v1.44.0` follows from grpc rather than being a third bump. I did not independently confirm those are the *minimum* versions that clear the advisories.

### Not checked

Meilisearch behaviour against a live server (no instance available — `Searcher` is covered only through the `clientAPI`/`indexAPI` fakes, which is a reasonable choice but means the generated filter and sort strings are never parsed by Meilisearch itself), `CloudRunTrigger` against real Cloud Run, and `PubSubPublisher`/`PubSubSubscriber` against real Pub/Sub.

ManavA:
## Independent review, round 2 — `5434e0f`

> **On the pending rebase.** Reviewed at `5434e0f` (on main `05fd34a`). I was told the tip may move to a content-only rebase onto main `174c823`, so I replayed that rebase locally: the result's diff against `174c823` is **byte-identical** (7,221 lines both ways) to `5434e0f`'s diff against `05fd34a`, and the full gate is green on the rebased tree as well. This verdict therefore stands for either tip. The only thing `174c823` adds is 3 lines in `ci.yml` — `pull-requests: read`, the correct fix for gitleaks-action's 403 — and the `secrets` job already carries `fetch-depth: 0` with a comment noting that a shallow clone would silently scan almost nothing. That is the right instinct, and it closes the one CI item I said I was taking on trust.

**Verdict: APPROVED.** All three blockers are fixed, and I confirmed each by running code rather than reading the diff. Every should-fix I raised is addressed. Four new non-blocking items below, two of which are worth a follow-up commit — but none of them justify holding this.

The two mutations that survived round 1 are now pinned by tests with names that say what they protect, which is the part I care about most.

### Gate at this tip

`gofmt` clean · `go build` clean · `go vet` clean · `go test ./... -race` green across all 15 packages · `golangci-lint run ./...` @ v2.6.0 **0 issues** · `govulncheck` 0 reachable · `gitleaks` clean on working tree **and** all 8 commits of history · live Postgres tests pass against a real PG 16. 158 test funcs in these four trees, up from 121.

### Blockers — all closed

**B1 — panic recovery. Fixed and verified.** `runRecovered` (jobs/runner.go:64) converts a panic into an error via a named return; `Scheduler.Run` (jobs/scheduler.go:79) adds a second recover whose `done <- struct{}{}` sits inside the defer, so a panic in the ticker logic itself can no longer wedge `Run` forever — that is the right second-order catch and the comment explains it correctly. Probe, under `-race`, panicking entry alongside a healthy sibling on 20ms tickers:

```
panicking entry ran 11 times; healthy sibling ran 11 times
terminal line for the panicking job: level=ERROR status=failed ok=false
```

No crash, no deadlock, `Run` returned. Panic value and stack both in the log. The sibling's next tick ran — which is the specific thing that failed last round.

**B2 — `LogSender`. Fixed and verified.** Default output now carries neither the address nor the model:

```json
{"level":"INFO","msg":"mail not sent: no provider configured",
 "template":"password-reset","to_hash":"ffbe8cff4f9f"}
```

With `LogBodies: true` and an Info-level handler, the body line is correctly absent — it is genuinely Debug-gated, not just labelled. See N3 below on the hash itself.

**B3 — Meili range-op quoting. Fixed.** All four range ops now go through `meiliValue` (search/meili/filters.go:14-21), and `TestBuildMeiliFilters_RangeOpsQuoteAHostileStringValue` holds the line — I removed the quoting again and it went red.

### Should-fixes — all closed

| # | Item | Verified |
|---|---|---|
| S1 | `CheckTemplates` 404/422 = missing, 401/403/429/5xx = cannot-check | mail/templates.go:27-41, with the status and body in the error; alias `url.PathEscape`d |
| S2 | Shared `events.Marshal` | events/publisher.go:22, used by both implementations |
| S3 | At-most-once documented, drops counted + logged | events/subscriber.go:26-37 and :61-69 — this is now one of the clearest docs in the repo; `DroppedCount()` added |
| S4 | `Searcher` honours ctx | `ctx.Err()` checked in all 8 methods, and per-page inside `PruneStale`/`RemoveDocuments` loops |
| S5 | Subpackage split | reproduced: `events` **1** non-stdlib dep (from 299), `jobs` **1** (from 24), `search` **31** (from 62). Compile-time assertions present: `var _ events.Publisher`/`Subscriber`, `var _ jobs.Trigger`, `var _ search.Index` |
| S6 | `Idempotent(ctx, g, job, key, fn)` | two different jobs sharing key `2026-09-17` both run now |
| S7 | tsvector doc | search/doc.go:11-17 now says "whole-lexeme matching, not substring matching… unstemmed" — which matches exactly what I measured against real Postgres last round (`bungalow` hits, `bung` does not, `charm` does not stem). `TextSearchConfig` added with a default helper |
| S8 | `Offset` clamped | search/postgres_query.go:85-90 |
| S9 | 22P02 no longer echoes the row | `wrapPgError` (search/postgres.go:357) |
| S10 | Both round-1 survivors pinned | `TestBuildPostgresQueries_DefaultsToOrderByID` + `TestPostgresIndex_Search_DefaultsToOrderByID_EndToEnd`; `TestPostgresIndex_PruneStale_ReturnsTheActualDeletedCount`. Re-applied both mutations: **both now killed** |
| S11 | alias escaping, style, goconst | done; lint 0 at the pin |

I also mutated the new code: removing the Offset clamp, the recipient hashing, and the range-op quoting each go red immediately.

### New, non-blocking

**N1 — the drop `Warn` is not rate-limited.** One burst of 5,000 publishes into a stalled subscriber:

```
dropped counter = 4983
WARN lines emitted = 4983, log bytes = 700981
```

700KB of log from one stalled subscriber. The round-1 fix for silent drops is right in principle, but "one Warn per dropped message" turns a degraded subscriber into a log-volume incident. Suggested: log the first drop per topic, then sample — every Nth, or once per interval — and lean on `DroppedCount()` for the exact number, which is what it is there for.

**N2 — `Publish` holds `b.mu` across the `Warn` write.** The logging call sits inside the `for` loop under `defer b.mu.Unlock()` (events/subscriber.go:104-114), so a slow log handler serialises publishing for *every topic*. Measured with a 2ms-per-write handler:

```
a publish to an UNRELATED topic waited 3.36ms while another topic's drops were logged
```

This is a small regression from the fix itself — round 1 held the lock but did no I/O under it. Suggested: collect the dropped topics into a local slice inside the critical section and log after `Unlock`. Fixing N1 largely mitigates N2, so one change likely closes both.

**N3 — `to_hash` is a bare, unsalted SHA-256 prefix, and the doc calls it "non-reversible".** I recomputed `sha256(strings.ToLower(strings.TrimSpace(to)))[:12]` independently and it matched the logged value exactly. For a **dev sender this is the right trade** and I would not change the mechanism: the property that matters is "no plaintext address in the log sink", and that is delivered. But an email address is a low-entropy, enumerable value, so anyone holding a candidate address confirms its presence by computing the digest — that is a dictionary lookup, not a preimage attack. The word to fix is in mail/logsender.go:64-66: "non-reversible" should read something like "does not contain the address in plain text; note that a holder of a candidate address can confirm a match by computing the same digest, so this defends against incidental disclosure, not against a determined reader."

I am flagging rather than blocking on this, even though I blocked on a doc-overclaims-a-safety-property in round 1 (B3). The distinction: B3's overclaim concealed a live injection path, whereas here the mitigation that actually matters is delivered and the overclaim concerns residual strength on a development-only sender. Worth a one-line doc change, not another round.

Two smaller notes on the same code: the 12-hex truncation is 48 bits, so correlation is reliable but not collision-free at large volume — fine as documented, worth knowing. And `wrapPgError` returns a fresh error rather than wrapping with `%w`, so `errors.As` can no longer reach the `*pgconn.PgError`; that is defensible as deliberate containment, just be aware it is a behaviour change for callers who were inspecting it.

**N4 — `Idempotent`: empty job name is accepted, and the `\x00` separator is ambiguous.** Probed both:

- `Idempotent(ctx, g, "", "2026-09-17", fn)` is **not refused** — it runs, and registers `"\x00" + key` as a namespace. Two independent callers that both omit the job name silently share a namespace, which is the bug the parameter was added to prevent. Cheap fix: return an error on an empty `job`.
- `(job="a\x00b", key="c")` and `(job="a", key="b\x00c")` map to the same guard key — verified, the second call was wrongly skipped. Exotic, since job names rarely contain NUL, but a length prefix or rejecting `\x00` in `job` closes it for one line.

### Not checked

Meilisearch against a live server (still fakes-only, which remains a reasonable call — but the generated filter and sort strings are still never parsed by Meilisearch itself), Cloud Run, and Pub/Sub. `gitleaks` run locally over the full history is clean, which is the part I can verify directly; the CI permissions fix is covered in the note at the top.

Nice work on the turnaround — particularly the `Subscriber` delivery-semantics doc and the `buildPostgresQueries`/`pgConn` split, which is what made the two survivor mutations testable.

PR #2, opened 2026-09-18
feat: add auth, admin, and textpolicy packages

ManavA:
## Summary

Three packages extracted from a production service's generic auth/admin/text-guard infrastructure, each usable on its own per ARCHITECTURE.md's layering rule (auth/admin at layer 4, textpolicy at layer 2 — auth and admin do not import each other).

- **`auth/`** — one `auth.Service` composable across identity sources:
  - `SourceLocal` (default): email/password (bcrypt, cost pinned by test; length bounded at both ends — 8 char minimum, 72-byte maximum matching bcrypt's own silent-truncation limit), signup duplicate → 409, email verification tokens, password reset tokens, account deletion, logout. Zero `Options{}` value works over in-memory storage.
  - `SourceFirebase`: `IDTokenVerifier` behind `FirebaseVerifier`, typed-nil-safe `SetIDTokenVerifier` guarded by a mutex, sanitized-display-name + logged-drop pattern for auto-registration.
  - `SourceOIDC`: generic OIDC (`OIDCVerifier`) — discovery with RFC 8414 issuer validation and same-origin `jwks_uri` enforcement, HTTPS-only issuers unless explicitly opted out for local development, JWKS fetch/cache with rate-limited and singleflighted on-miss refetch, RS256-only, required expiry, exact (not merely contains) audience match. Preset issuer/audience table for Auth0, Google, Apple, Cognito, Clerk — no vendor SDKs.
  - Sources compose: Local and Firebase/OIDC link by verified, case-normalized email only.
  - Sessions: opaque (default, revocable, bounded with a periodic eviction sweep) or JWT (must be opted into explicitly, since it cannot be revoked), each carrying a required, package-specific `aud` claim so an auth token cannot be replayed as an admin token or vice versa.
  - `UserStore`/`VerificationStore`/`PasswordResetStore`/`SessionStore` interfaces; in-memory implementations plus **`auth/pg`**, a Postgres-backed implementation of all four (citext case-insensitive email, tested against `pg/testdb`, including a migration replay test).
  - Per-IP rate limiting via `httpx/middleware.RealIP` + `RateLimit`; handler logging via `httpx.Logger(ctx)`; a real store failure surfaces as 503 and gets logged rather than being reported as an ordinary 401.

- **`admin/`** — its own JWT session implementation (can't import `auth`), a `chi.Router` skeleton on `httpx/middleware` with its own CORS origin, a `Seed` function plus a runnable `admin/cmd/seed` (password via env/stdin, `-force` guard against a populated table, now built on **`admin/pg`**), and `CheckNoForbiddenFields`.

- **`textpolicy/`** — `Policy`/`Rule`/`Check`, normalization that strips every Unicode control and format character plus the complete `Default_Ignorable_Code_Point` property (4,174 code points, transcribed from Unicode 18.0.0's own data — not filtered down to only the entries with a named, assigned category, since several of its ranges are reserved/unassigned and a category-only predicate silently drops exactly those), then folds width and compatibility variants — stripping before folding is what keeps normalizing already-normalized text a no-op. `CheckStored` checks and returns the normalized text, since what gets checked must be what gets stored. `Joined`, and an obviously-fake `ExamplePolicy`.

## Fixes since the previous push

**A real gap in the stripped set, found on the rebased tree.** The predicate that decides what normalize strips had been narrowed to "every control character, every format character, plus a small hand-verified addition" — which is a good description of most of `Default_Ignorable_Code_Point`, but not all of it: several of that property's ranges are `Cn` (reserved/unassigned — held invisible so that whatever character Unicode eventually assigns there stays invisible by default), and an unassigned code point matches no category table at all, format category included. The generator that produced the "small addition" had, in an earlier pass, filtered out exactly the entries that were already format characters, which silently dropped every `Cn` entry along with them — nearly 3,800 code points, including `U+2065` and four large reserved blocks under `U+E0000`, stopped being stripped. `defaultignorable.go` now carries the complete property, transcribed directly from `DerivedCoreProperties.txt` line by line rather than filtered by category, and `shouldStrip` ORs it together with `unicode.Cf`/`unicode.IsControl` rather than replacing them — a superset by construction, not a hand-curated one. `defaultignorable_test.go` adds a second, independently transcribed copy of the same property and checks the production table against it, plus a dedicated test that `shouldStrip` covers every code point in that independent copy — the actual guarantee that matters, checked against a source the production table cannot influence.

**`golangci-lint` was clean only with a warm cache.** Reproduced from a cold one (`golangci-lint cache clean` first) and found 4 issues: two `goconst` findings for the literal `"sessions_revoked"` repeated across `Logout`, `ResetPassword` and `DeleteAccount` (now a package constant, `sessionsRevokedKey`), and two `gosec` G704 (SSRF via taint analysis) findings on the JWKS fetch in `auth/oidc.go`. The gosec pair are false positives — `v.jwksURL` is either a caller-configured static value or a value `discoverJWKSURL` already required to share the configured issuer's origin before it was ever stored — confirmed by removing the same-origin check and watching `TestNewOIDCVerifierRejectsCrossOriginJWKSURI` go red; both call sites now carry a `//nolint:gosec` with that reasoning inline rather than a bare suppression.

**Should-fix items, all addressed:** `Normalize`'s idempotency depended on an `x/text/unicode/norm` release that truncated one code point mid-decomposition; bumped `golang.org/x/text` to v0.42.0 (and `golang.org/x/crypto` to v0.56.0, closing two more advisories in the security scan). `ARCHITECTURE.md`'s package map and import-layer list now actually include `auth`, `admin`, `textpolicy` and their `pg`/`cmd/seed` subpackages, rather than the "three packages planned" placeholder the previous push left behind. Three items carried from an earlier push were fully implemented but never pinned with their own test: the 72-byte password ceiling (now has a boundary test at exactly 72 and 73 bytes, plus a handler-level test), `CheckStored`'s rune-vs-byte length counting (now has a test using a 5-rune, 10-byte string that a byte limit would wrongly reject), and the in-memory session store's periodic eviction sweep (now has a test that creates `sweepInterval` already-expired sessions, confirms they are still present immediately before the threshold write, and confirms they are gone immediately after it). All three were mutation-checked (see below).

**Confirmed correct, no change needed:** the migration-replay coverage ask from an earlier push turned out to already be satisfied by this repository's own `pg/migrate` test, which discovers every `*/migrations` directory under the module root and replays each one — a stronger guarantee than a per-package test, since it cannot be forgotten for a package added later. `auth/pg` and `admin/pg` keep their own replay tests as well, which is redundant with that gate but not wrong. Case-insensitive account linking is handled at the handler boundary (email is normalized before every store call); a lower-level divergence in the in-memory store's own comparison is unreachable through the public API and is called out in "Left out" below rather than fixed, since a fix there would be defending against a path that does not exist. The `citext` schema-qualification fix from an earlier push was re-verified as a genuine bug, not a false alarm.

## Mutation checks (mutate, watch red, revert)

| File | Mutation | Result |
|---|---|---|
| `textpolicy/defaultignorable.go` | Narrowed `shouldStrip` back to `IsControl \|\| Cf` only (the actual regression) | Red — `TestNormalizeStripsNamedBypassCharacters`'s `Cn` subtests (`U+2065`, the `U+E0000`/`U+FFF0` reserved ranges), `TestShouldStripIsSupersetOfDefaultIgnorable` |
| `auth/oidc.go` | Removed the discovery-document same-origin `jwks_uri` check that makes the JWKS fetch's `//nolint:gosec` justified | Red — `TestNewOIDCVerifierRejectsCrossOriginJWKSURI` |
| `auth/handlers.go` | Disabled the 72-byte password ceiling | Red — `TestSignupRejectsPasswordOverBcryptsLimit`, `TestValidatePasswordLength/one_over_the_maximum` |
| `textpolicy/guard.go` | Swapped `CheckStored`'s rune-count length check for a byte-count one | Red — `TestCheckStored/the_length_limit_counts_runes,_not_bytes` |
| `auth/session_store.go` | Disabled the periodic sweep (`maybeSweepLocked`'s threshold check) | Red — `TestMemorySessionStoreSweepEvictsExpiredSessions` (map size stayed at 127 instead of collapsing to 1) |

All reverted; suite green afterward. Earlier pushes' mutation tables (session-token audience claims, OIDC discovery hardening, session revocation, migration `CASCADE`, `ON CONFLICT` token-collision handling, non-positive TTL rejection, store-failure-vs-401, idempotency ordering, `CheckStored` normalization) are unchanged and still green; see the commit for detail.

## Test counts

209 top-level `Test` functions across the three packages (auth 119, auth/pg 29, admin 36, admin/pg 8, textpolicy 17) — roughly 320 including table-driven subtests.

`go build/vet/test -race ./...` (all packages, including the repository's own migration-replay gate), `golangci-lint run ./...` against a cold cache (0 issues), `govulncheck ./...` (0 in-code vulnerabilities; the remaining imported/required advisories are in `grpc`, `fasthttp`, `x/image` and `x/crypto/openpgp` — none reachable from code these three packages own), and `gitleaks detect` (clean) all pass.

## Left out, and why

- **In-memory store hard caps with oldest-first eviction**: the periodic TTL sweep is in place and now pinned by a test, but there is no hard cap independent of expiry — a store fed sessions/tokens with very long TTLs faster than they expire can still grow unbounded. Needs a policy decision (per-user vs. global cap, eviction order under contention) this scope shouldn't make unilaterally.
- **Postgres store `PurgeExpired` and an `expires_at` index**: not added. Needs a `jobs` package entry to be worth doing at all, and this scope doesn't own `jobs`.
- **In-memory store's own case-sensitivity at the storage layer**: handler-level normalization (trim + lowercase before every lookup) already makes this unreachable through the public API; a store-layer fix would be defending against a path that does not exist rather than a real one.
- **`admin/cmd/seed` hardening beyond what shipped**: role allowlist validation, a transaction-plus-unique-constraint guard against a duplicate-email race between the existence check and the insert, and dedicated tests for the seed command itself. What shipped: password via env or a non-echoing stdin prompt, a `-force` guard against seeding a populated table, and a pinned bcrypt cost.
- **OIDC revocation check**: no standard equivalent exists for generic OIDC; documented on `OIDCVerifier`.
- **`SessionJWT` revocation**: an honestly-documented, explicit-opt-in no-op — implementing it would need the same storage `SessionOpaque` already provides, at which point there is no reason to use `SessionJWT` at all.

Do not merge — tip is `1e900e9` (adds a second follow-up commit pinning golangci-lint-action to v2.9.0, the earliest release built with a go1.26 toolchain — confirmed green on the actual GitHub Actions run, not just locally).


ManavA:
## Independent review — CHANGES REQUESTED

Reviewed at `530dffd`, in a clean clone, as an attacker. I re-applied all eight mutations from the PR description and added seventeen of my own; every finding below is backed by a probe I ran, not by reading.

Credit where it's due first, because these were the hard parts and they hold up:

- **Algorithm confusion is genuinely blocked.** `alg: none` and `alg: HS256` signed with the RSA public key are both rejected by `OIDCVerifier`. The PR's note that `WithValidMethods` is redundant with the keyfunc's `*rsa.PublicKey` return is **correct** — I reproduced the still-green result. Disclosing that instead of hiding it is the right call.
- Unverified federated email never links to an existing account (fail-closed, confirmed by mutation T4).
- Tokens are hashed at rest, single-use, expiring, 32 bytes from `crypto/rand`. Verification and reset stores are correctly separate.
- Rate limiting keys on `r.RemoteAddr`, not a spoofable `X-Forwarded-For`, with the trust model stated.
- Wrong `aud` (string *and* array), wrong `iss`, and a JWKS 500 mid-rotation all behave correctly — the last falls back to the cached key rather than failing the request.
- Zero `Options{}` really does produce a working local-only service. `gitleaks detect` is clean over the tree and all 6 commits. `go build`, `go vet`, `gofmt`, `go test -race` all clean. Test counts (85/29/6) are accurate.

### Blockers

**1. An end-user session token authenticates as an operator.** `auth.sessionIssuer` and `admin.sessionIssuer` emit byte-identical tokens: HS256, claims `{sub, iat, exp}`, no `aud`, no `iss`, no type discriminator. Nothing in either package's docs warns against configuring both from one `JWT_SECRET`, which is the obvious thing to do.

```
end-user auth session token presented to admin.RequireAdmin
  -> status=200  reached_handler=true  adminID="user_1"
```

That is full privilege escalation into every route mounted the way `admin.Router`'s own doc comment shows. Fix: give each package a required, validated distinguishing claim (`aud: "admin"` / `"user"`), and reject a token that lacks it.

**2. `OIDCVerifier` accepts a token with no `exp` claim.** `jwt.Parse` does not require expiry unless you ask it to.

```
token with NO exp claim -> err=<nil>  identity={UID:no-exp-user EmailVerified:true}
```

A token that never expires, from any issuer that omits `exp`. Add `jwt.WithExpirationRequired()`.

**3. `textpolicy` normalisation strips 262 of ~4,174 default-ignorable code points.** The listed ones work; the ones an attacker reaches for second do not.

```
U+200B ZWSP  (doc claims stripped)     blocked=true
U+FEFF BOM   (doc claims stripped)     blocked=true
U+200E LRM                             blocked=false
U+200F RLM                             blocked=false
U+202E RLO                             blocked=false
U+2066 LRI                             blocked=false
U+034F CGJ                             blocked=false
U+180E MVS                             blocked=false
U+3164 HANGUL FILLER                   blocked=false
U+2061 FUNCTION APPLICATION            blocked=false
```

This is the exact evasion the package exists to stop: invisible, present byte-for-byte, and `\b` treats each as a token boundary. `normalize.go`'s argument for an explicit list over a category test is sound — the list is just incomplete. Add at least `U+061C`, `U+200E–U+200F`, `U+202A–U+202E`, `U+2061–U+206F`, `U+034F`, `U+180B–U+180F`, `U+115F–U+1160`, `U+17B4–U+17B5`, `U+3164`, `U+FFA0`, `U+FFF0–U+FFF8`, and widen `U+E0100–U+E01EF` to `U+E0000–U+E0FFF`. The existing tests only exercise code points that already pass, so they cannot catch this class.

**4. Nothing can revoke a session.** There is no `/logout` route; `SessionStore.Revoke` is declared, implemented, and **never called anywhere in the package**; and a password reset leaves every existing session live:

```
AFTER password reset, stolen session validate -> userID="user_1" err=<nil>
POST /logout -> 404   POST /signout -> 404   POST /revoke -> 404
```

`SessionOpaque` is the default specifically because it is revocable, and that property is unreachable through the Service's own API. A stolen token survives the victim's password reset for the full 7-day TTL. `DeleteAccount` has the same hole. Fix: add `RevokeForUser(ctx, userID)` to `SessionStore`, call it from `ResetPassword` and `DeleteAccount`, and mount a logout route.

**5. OIDC discovery follows `jwks_uri` to any host, and never validates the document's own `issuer`.**

```
discovery jwks_uri -> foreign host; NewOIDCVerifier err=<nil>, foreign host hit 1 time
token signed by the FOREIGN key -> err=<nil>
```

`discoveryDocument.Issuer` is decoded and never compared to the configured issuer — RFC 8414 §3.3 requires that comparison. Combined with an unconstrained `jwks_uri`, a hostile or compromised discovery document redirects trust to a key the attacker controls, and the fetch itself is an SSRF primitive against link-local and internal addresses. Also: neither response body is size-limited, and the default `http.DefaultClient` has no timeout.

**6. The documented Auth0 preset rejects valid Auth0 tokens.** Auth0 issuers end with `/` and the `iss` claim carries it; the package doc's preset table says so. `NewOIDCVerifier` then does `strings.TrimRight(opts.IssuerURL, "/")` and `jwt.WithIssuer` compares exactly.

```
issuer configured ".../", token iss ".../" -> token has invalid issuer
```

One of the five advertised providers does not work as documented. Compare against the discovery document's `issuer` (which fixes #5 too), or accept both forms.

### Should fix

**7. Unknown-`kid` refetch storm.** 20 unauthenticated bogus tokens caused 20 outbound JWKS fetches — one per request, no negative cache, no backoff, no single-flight. The keyfunc runs *before* signature verification, so this costs an attacker nothing. Add a minimum interval between refetches and collapse concurrent ones.

**8. Account linking is case-sensitive, so it silently doesn't link.** `Signup` lowercases; `resolveIdentityUser` passes `identity.Email` through raw.

```
local account     id=user_1 email=alice@example.com
federated identity id=user_2 email=Alice@Example.com  <- second account, same mailbox
```

The feature's whole purpose is not creating a second account for the same person. `auth/pg/migrations/0001_auth.sql` bakes the same problem in with a case-sensitive `TEXT UNIQUE`.

**9. A long passphrase is a 500.** bcrypt refuses input over 72 bytes; `Signup` maps that to `500 invalid request` rather than a 400 naming a maximum.

**10. In-memory stores never evict.** 500 sessions created with a 1ns TTL, all expired: 500 entries still held. Unbounded growth in the default configuration, driven by unauthenticated `/login` traffic.

**11. `CheckStored`'s `maxLen` is bytes and says nothing about it.** `CheckStored(10 CJK chars, maxLen=10)` → too long. A caller sizing the limit from a `varchar(N)` column silently rejects legitimate CJK at a third of the advertised length. One word in the doc fixes it.

**12. Homoglyphs are out of scope but the doc implies otherwise.** `doc.go:31-33` — "robust to encodings nobody has thought of yet" — is not true of Cyrillic `а`, and after #3 it is not true of invisibles either. Needs a Limitations section and that sentence removed.

**13. `sanitizeDisplayName` produces invalid UTF-8.** `cleaned[:200]` is a byte slice; 100 CJK runes → `valid UTF-8 = false`. Zero-width characters also survive it, because `unicode.IsControl` is Cc only and these are Cf — while this same PR ships a package that strips exactly those.

**14. `admin/cmd/seed` takes the password as a command-line flag.** argv is world-readable via `ps`, and it lands in shell history. Read it from stdin or the environment. Seed also has no guard against running against a populated table, no password floor (`"a"` is accepted), and no role validation.

**15. No minimum length on `Options.Secret`** in either package — a one-character HS256 signing key is accepted.

**16. The PR is based on `94c819b`, not current `main`.** `pg`, `httpx`, `config` and `log` have since landed, so two "hasn't landed yet" justifications are now stale, and `go.mod`/`go.sum` conflict on merge. Concretely, after rebase: `pg/migrate` requires an `.up.sql` suffix, so `0001_auth.sql` and `0001_admin.sql` would not be picked up by keel's own runner; and `httpx/middleware.RealIP` already solves the trusted-proxy problem that `keyByRemoteAddr`'s comment tells the caller to solve themselves — point at it by name.

**17. Signup enumeration via 409 is real and undocumented.** The body is literally "an account with this email already exists". That is a defensible trade-off, but `doc.go`'s "Error responses" section only covers the 401 case, so the trade-off isn't recorded where a reader would look.

### Mutation table

Your eight, re-applied by me — all reproduce exactly as described, including the honest still-green one:

| Mutation | Claimed | Observed |
|---|---|---|
| `oidc.go` drop `WithValidMethods` | still green | **still green** — confirmed, keyfunc type is the real guard |
| `oidc.go` drop `WithAudience` | red | red — `TestOIDCVerifierRejectsWrongAudience` |
| `service.go` typed-nil always false | red | red — 3 tests incl. `TestIsNilInterfaceValue` |
| `handlers.go` drop `EmailVerified` link guard | red | red — `TestExchangeIDTokenFailsClosedOnUnverifiedEmailCollision` |
| `handlers.go` disable control stripping | red | red |
| `guard.go` disable length check | red | red |
| `fields.go` skip own-level keys | red | red — 2 tests |
| `router.go` CORS hardcoded `*` | red | red — 2 tests |

Mine, testing whether the suite pins the controls it relies on:

| Mutation | Result |
|---|---|
| `oidc.go` drop `WithIssuer` | red |
| verification/reset: don't mark consumed | red (both) |
| verification/reset: ignore expiry | red (both) |
| store raw token instead of its hash | red |
| `RequireAuth` / `RequireAdmin` accept any bearer prefix | red (both) |
| Signup skips password-length check | red |
| forgot-password 404s on unknown email | red |
| **`oidc.go` keyfunc returns an HMAC secret for HS256** | **STILL GREEN** |
| **opaque session `Validate` ignores expiry** | **STILL GREEN** |
| **opaque session token 32 bytes → 4 bytes** | **STILL GREEN** |
| **`auth`/`admin` `ValidateToken` drop `WithValidMethods`** | **STILL GREEN** (both) |
| **bcrypt `DefaultCost` → `MinCost`** | **STILL GREEN** |

The first of those is the one that matters. Your note argues `WithValidMethods` is redundant because the keyfunc only ever returns an `*rsa.PublicKey` — true today, and **nothing in the suite pins it**. I made the keyfunc hand back an HMAC secret for HS256 and the suite stayed green. The two controls that together prevent algorithm confusion are each individually untested, so whichever one a future change removes, CI will agree. Please add explicit `alg: none` / `alg: HS256` rejection tests to `OIDCVerifier` *and* to both `sessionIssuer`s, plus an expiry test for opaque sessions.

### Style and hygiene

**Blocking on the owner's "no domain" rule** — CONTRIBUTING.md says nothing in this repository may carry a domain, and these are the leaks:

- `auth/handlers.go:274` — "(GoReal issue #297: an unsanitized display name from a first sign-in reached the users table unchanged)". The observation is right and should stay; the citation is unresolvable from this repo and names the private project keel was extracted from. Drop the issue number, keep the sentence. This is the only such reference in the PR.
- `admin/fields_test.go:12` — "a public view and a **broker**/operator-only view"; `admin/fields_test.go:52` — `InternalNotes: "do not show **buyers**"`. Real-estate vocabulary in a fixture whose own comment advertises itself as a generic stand-in. Use "operator-only" and "not for public display".
- `admin/doc.go:31-32` cites "showing instructions" as its worked example — an MLS term. Pick a second neutral example.

Other style findings, all minor:

- `auth/verification_handlers.go:86` — "rather than a cheerful 200" — jokey personification; the sentence works without it.
- `auth/verification_handlers.go:15-16` and `auth/service_test.go:107-108` both end a good, concrete sentence with "the defect/bug class this package exists to prevent". The package exists to provide authentication; the preceding clause already names the failure. Delete the trailing clause in both.
- `admin/fields_test.go:13-15` is backwards: it says `assertNoProtectedFieldFor` is "the pattern this package exports as `CheckNoForbiddenFields`", but it is a local wrapper *around* the exported function.
- `auth` uses receiver `h` on a type named `Service` across 16 methods; `admin` uses `s` for the identical type. `h` is a leftover from a `Handlers` type. Rename to `s`.
- `auth/password_reset_handlers.go:110` — `DeleteAccount` lives in the password-reset file. `auth/account_test.go` already knows where it belongs; move it to `auth/account.go`.
- `auth/password_reset_handlers.go:39,97` bake fixed end-user copy into the library with no way to override or localise it. Either say so in the doc or put them on `Options`.
- `auth/service.go:236` writes `h.idVerifier` unsynchronised from an exported setter, while `ExchangeIDToken` reads it from request goroutines. `-race` is clean because every test wires before serving, so this is latent, not found — but nothing in the signature says "before you start serving". Document it or guard the field.
- `auth/password.go:28` and `admin/handlers.go:42` run a full `bcrypt.GenerateFromPassword` at cost 10 during package init, so any binary that merely links these packages pays ~50–100ms of startup for a feature it may never use. `sync.OnceValue` on first login miss gets the same property for free.
- `auth.ErrInvalidToken` and `admin.ErrInvalidToken` are exported but returned only by unexported types and swallowed by the middleware — no caller can reach them. `Identity.Picture` is populated by both verifiers and then discarded: never copied to `User`, no column in the migration, absent from `SessionResponse`.
- `auth/session.go:45` `IssueTokenWithExpiry` has no production caller, and its doc justifies itself with "an admin session, for instance" — but `admin` architecturally cannot use this type, as `admin/session.go:22-25` explains.

### Docs

**ARCHITECTURE.md, README.md and CONTRIBUTING.md were not touched by this PR** (the diff is empty). Mostly defensible since all three pre-declared these packages, but three things are now out of step:

- The Layout tree omits `admin/cmd/seed/`, `auth/pg/migrations/` and `admin/pg/migrations/`. Also, `auth/pg/` and `admin/pg/` read like "auth's dependency on `pg`", the exact edge the layering forbids — consider `auth/migrations/`, or state the schema-only convention.
- ARCHITECTURE.md says a function that takes a `context.Context` and ignores it "is worse than one that does not take it at all". Every `Memory*Store` method does exactly that (`_ context.Context`, ~15 times). That is correct Go for an in-memory implementation, but the rule as written admits no exception — add the carve-out.
- CONTRIBUTING.md says tests are table-driven; 2 of 114 are. Relatedly, the `.golangci.yml` comment justifies the `goconst` exclusion by "table-driven tests repeat the same literal across cases", but the literal that actually trips `goconst` is `"hunter2hunter2"` repeated across *separate test functions*. The exclusion is fine; the reason attached to it doesn't describe this code, which matters more than usual in a repo whose README says a decision with its reason attached survives refactoring.

### Verified clean

`go build ./...`, `go vet ./...`, `go test -race -count=1` (all three packages), `gofmt -l`, and `gitleaks detect` over the tree and all 6 commits — all clean. `golangci-lint` at the repo's pinned v2.6.0: **0 issues**. `govulncheck ./...`: **0 reachable**. `go.mod` pins exactly `go 1.25.0`.

Import layering verified mechanically over `Imports`, `TestImports` and `XTestImports`: the three new packages import **no** keel package at all, and the repo's only internal edge is `admin/cmd/seed → admin`. `admin` does not import `auth` and `auth` does not import `admin` — though note the sibling rule is satisfied trivially here, since levels 1–3 don't exist on this branch. The deliberate JWT-wrapper duplication is documented at `admin/session.go:22-25`, which is the right place for it.

One dependency note: `golang.org/x/crypto@v0.53.0` is a **direct** require introduced by this PR, and sits behind three advisories with fixes available (GO-2026-6303 → v0.55.0; GO-2026-6354/6355 → v0.56.0). `govulncheck` stays green because bcrypt isn't the affected surface, so this is hygiene rather than a blocker — but it means the green result is doing less work than it looks.

`CheckNoForbiddenFields` recurses correctly through nested objects and arrays and its error does not echo values. The match is case-sensitive — `Private_Remarks` and `PRIVATE_REMARKS` both pass a `private_remarks` allowlist — which is defensible for checking your own encoded output, but say so in the doc.

`.golangci.yml`: the `goconst`-on-`_test\.go` exclusion names one linter, is scoped to tests only so it cannot leak into production code, matches the existing `errcheck`/`gosec` entry's shape, and the diff touches nothing else. **Acceptable as written** (modulo the inaccurate reason noted above).

ManavA:
## Round 2 — CHANGES REQUESTED

Reviewed at `4bcfb62`. **All six round-1 blockers are genuinely fixed**, and I verified each by probe rather than by reading. The new work is substantial and most of it is right. But the textpolicy fix closed the hole I named and left the surrounding class open, and the table the whole fix rests on is guarded by two tests that cannot fail — I corrupted it and the suite stayed green.

### Round-1 blockers: all six verified fixed

| # | Round-1 blocker | Verified at `4bcfb62` |
|---|---|---|
| 1 | auth/admin JWTs interchangeable | **Fixed.** `aud` `keel:auth`/`keel:admin` required and validated; **401 in both directions**, handler not reached |
| 2 | OIDC accepted a token with no `exp` | **Fixed.** `exp claim is required` |
| 3 | textpolicy stripped 262 of ~4,174 ignorables | **Fixed for the property** — 0 bypasses across all 4,174; U+200E/200F/202E/2066/034F/180E/3164/2061 all blocked |
| 4 | No session revocation anywhere | **Fixed under opaque.** `Logout` + `RevokeAllForUser` wired into reset and delete; stolen session dies on reset |
| 5 | OIDC discovery SSRF | **Fixed.** Foreign-host `jwks_uri` refused, scheme mismatch refused, issuer compared exactly, 1 MiB caps, 10s timeout |
| 6 | Auth0 trailing-slash preset broken | **Fixed.** Issuer preserved for `iss`, trimmed only for the discovery URL; Auth0-shaped token verifies |

Round-1 should-fixes also confirmed fixed: unknown-`kid` storm (50 bogus tokens → **1** fetch), >72-byte password → **400**, eviction (5,000 expired → **8** retained, live session survives 1,000 churn writes), `sanitizeDisplayName` rune-safe and strips invisibles, secret minimum 16 bytes, `SetIDTokenVerifier` mutex-guarded with a matching read accessor, `Identity.Picture` removed, `RealIP`/`RateLimit` now taken from `httpx/middleware` with a secure zero value, migrations renamed to `.up.sql`/`.down.sql`, the `GoReal issue #297` reference and the `broker`/`buyers`/`showing instructions` domain vocabulary all gone. `gitleaks` clean over 11 commits; `golangci-lint` **0 issues**; `govulncheck` 0 reachable; `go vet`, `gofmt`, `go test -race -count=1` all clean.

### Blockers

**1. `textpolicy.Normalize` is not idempotent, and that is a live bypass.** The order is trim → NFKC → strip → collapse. A default-ignorable code point placed *between a base letter and its combining mark* blocks NFKC composition; stripping it afterwards leaves a decomposed sequence that NFKC would have composed.

```
Normalize("cafe​́")  once="café"  twice="café"  idempotent=false
Normalize("ᄀ​ᅡ") once="가"    twice="가"     idempotent=false

rule (?i)\bcaf\x{00E9}   precomposed              blocked=true
                         decomposed               blocked=true
                         ZWSP between e and acute blocked=FALSE
                         CGJ  between e and acute blocked=FALSE
```

All four render identically. U+034F is *designed* to block composition and is one of the eight code points round 1 named — so the new table strips it while the deeper effect survives. **One-line fix: re-apply `norm.NFKC` after the strip.** A normaliser whose output is not normalised is a bug on its own terms, and `Normalize` is exported for callers to clean text before storage.

**2. Invisible characters outside `Default_Ignorable_Code_Point` still defeat every rule.** The fix treats that property as a synonym for "invisible". It isn't:

```
U+0001 SOH  blocked=false     U+0080 C1              blocked=false
U+001B ESC  blocked=false     U+FFF9 interlinear     blocked=false
U+007F DEL  blocked=false     U+0600 arabic num sign blocked=false
                              U+00A0 NBSP            blocked=false
```

`"ban\x01ned"` is invisible in essentially every renderer, passes the guard, and is handed back by `CheckStored` to be stored verbatim. U+FFF9–FFFB, U+0600 and U+13430 are *deliberately excluded* from the property by Unicode, so this is a category error rather than a gap the property can close. The repo already knows: `auth/handlers.go:308-319` has to do its own `unicode.IsControl` strip *before* it can trust `Normalize` — the display-name sanitizer is stronger than the security guard it wraps. Fold `unicode.IsControl` into `Normalize`, or document the boundary.

**3. The two tests guarding the ignorable table cannot fail.** I swapped one range for another of identical size (`0x1BCA0..0x1BCA3` → `0x1BCB0..0x1BCB3`), moving four genuine format characters out and four unrelated ones in:

```
go test -count=1 ./textpolicy/   ->   ok
```

The entire suite passes with a corrupted table, while an independent probe finds the four new bypasses immediately. `TestDefaultIgnorableTotalMatchesUnicode` compares the table's sum to `defaultIgnorableTotal`, a constant declared in the file it is testing. `TestNormalizeStripsEveryDefaultIgnorableCodePoint` iterates the table and asserts `normalize` strips each entry — but `normalize` decides what to strip by calling `unicode.Is` on that same table. It tests the table against itself. Neither can detect the drift they are named for.

Compounding it: **no generator is committed** (no `go:generate`, no generator program, no Makefile target), and the "pin" is prose — the comment names Unicode 18.0.0 but the URL is `/Public/UCD/latest/ucd/`, which means "re-parse whatever Unicode ships that day". The table is *correct today* (I verified all 4,174 against UCD 18.0.0 and, independently, against Go's stdlib range tables — exact match both directions), but nothing in the repo can reproduce or re-verify it. Replace both tests with a stdlib set-equality cross-check (~20 lines, no network, actually fails on drift), commit the generator, and pin the URL to `/Public/18.0.0/`.

**4. `CheckStored` returns the raw string while having checked the normalised one, and the doc guarantees the opposite.** `guard.go:24-26` says it *"guarantees that whatever a caller writes to storage is exactly the value this call checked — never a value derived from it."* Both halves cannot be true:

```
input "‮dennab" -> checked as "dennab" -> err=<nil> -> STORED "‮dennab"
```

That stored string renders as **banned**. The guard evaluated `dennab`, passed it, and returned a value that displays as the word the rule forbids. Either return the normalised string, or reject input where `normalize(text) != text`, or drop the guarantee sentence — but as written the documented contract is false in the one case that matters.

### Should fix

**5. Case-insensitive linking is only fixed in the pg store; the default in-memory store still forks accounts.**

```
local=user_1 (alice@example.com)   federated "Alice@Example.COM" verified -> user_2
```

`resolveIdentityUser` (`auth/handlers.go`) passes `identity.Email` through raw — `Signup`, `Login` and `ForgotPassword` all lowercase, this path does not — so case-insensitivity rests entirely on citext. The result is two implementations of `UserStore` with **different security-relevant semantics** and an interface contract that says nothing about case, so the package's own tests against the memory store cannot catch a regression in the pg one.

`0001_auth_users.up.sql` states the assumption out loud — *"Handlers already lowercase an email before using it"* — and that is exactly the sentence that is not true of the federated path. citext is silently covering for a handler invariant one handler does not uphold; the memory store has no such backstop. Normalise at the package boundary as the other three handlers already do: that fixes both backends, makes the contract uniform, and makes the migration comment true.

**6. The `SessionJWT` no-op revocation is documented well but still lies to the client.** You asked me to judge this. The docs are genuinely good — `SessionMode` and `Logout` both state plainly that revocation is a no-op under JWT and the token keeps working. But:

```
JWT: ResetPassword -> 200 {"message":"Your password has been reset. You can sign in now."}
     pre-reset session after reset -> still VALID
JWT: Logout -> 204 No Content
     session after logout -> still VALID
```

A `204 No Content` is an unqualified success for something that did not happen, and the reset body affirmatively tells the user they are safe. Documentation does not reach the operator who picks `SessionJWT` for scaling reasons. I'd make it an explicit opt-in — `Options.AllowUnrevocableSessions bool`, refused otherwise — so the trade-off is acknowledged in code at the point of choice rather than in a doc comment someone may never open. At minimum, have `Logout` and `ResetPassword` report revocation status in the response.

**7. The OIDC keyfunc's type invariant is still unpinned — the one round-1 item the new pins don't cover.** Three of the five claimed pins are real (bcrypt cost, token entropy, opaque expiry all go red). The alg pins go green when `WithValidMethods` alone is removed, but that is the benign redundancy you disclosed honestly in round 1 — I confirmed the *behaviour* is pinned by removing both guarantees at once, and `TestSessionIssuerRejectsAlgNone` correctly goes red. The gap is the OIDC verifier. Making its keyfunc return a non-RSA key:

```
HS256 token forged with a NON-RSA key -> err=<nil> identity={UID:ATTACKER Email:attacker@evil.test EmailVerified:true}
suite: green
```

Full identity forgery, invisible to 204 tests. Pin it with a test that asserts the keyfunc only ever yields `*rsa.PublicKey`.

**8. Three more controls are unpinned.** Reverting the rune fix to a byte count (`textpolicy`), removing the >72-byte password check, and disabling the memory-store eviction sweep all leave the suite **green** — each is a round-1 fix that can now silently regress. `admin/seed.go`'s bcrypt cost is unpinned while `auth`'s is.

**9. Minor.** `aud` as an array containing *both* `keel:auth` and `keel:admin` is accepted by both validators (`jwt.WithAudience` is a containment check) — not reachable through either package, since neither mints a multi-audience token, but worth an exact-match assertion if a third issuer ever shares the secret. A plain-`http` issuer URL is accepted, which OIDC forbids. ARCHITECTURE.md / README.md / CONTRIBUTING.md are **still untouched** — the Layout tree now also omits `auth/pg`, `admin/pg` and `admin/cmd/seed`, and the Context-is-honoured and table-driven-tests conventions remain contradicted by their own first users. `golang.org/x/crypto` is still `v0.53.0`, behind three advisories with fixes (GO-2026-6303 → v0.55.0, GO-2026-6354/6355 → v0.56.0). "a cheerful 200" (`auth/verification_handlers.go:86`) and receiver `h *Service` across 19 `auth` methods (vs `s` in `admin`) both survive. The 409 signup-enumeration trade-off is still not recorded in `doc.go`. `textpolicy` still panics on a nil `*Policy` or a `Rule` with a nil `Pattern`, still does not document that rule matching is case-sensitive, and still treats a negative `maxLen` as unlimited. `Check` itself has no length bound, and because `maxLen` counts *raw* runes, a 192 KB body of U+FDFA costs ~67 ms CPU and ~37 MB allocated.

### The new Postgres stores and the seed command: clean

Audited separately and they hold up. **No dynamic SQL anywhere** — all 19 `Exec`/`Query`/`QueryRow` call sites in `auth/pg` and `admin/pg` use static query strings with bound parameters; the only `fmt.` calls in the tree are the seed command's terminal output. Verification, reset and session tokens are stored hashed. `CREATE EXTENSION IF NOT EXISTS citext` fails loudly on a role without privileges rather than falling back to a case-sensitive column (`IF NOT EXISTS` suppresses only the already-exists case), and `pg/migrate` runs each file in a transaction and stops naming it — reasoned from Postgres semantics, not tested against a restricted role.

The seed command addresses round 1 fully: the `-password` flag is gone in favour of `ADMIN_SEED_PASSWORD` or one line of stdin, there is a password-length floor, and a populated `admin_users` is refused without `-force`. To answer the question directly — **`-force` adds, it never overwrites**: it only bypasses the row-count guard, and a duplicate email still fails with `ErrDuplicateEmail`, so it cannot silently reset a sitting operator's password. It also now opens the pool through keel's own `pg.Open` rather than raw pgx, which retires the round-1 staleness note. (`-role` remains free-form.)

Worth calling out as a good decision: the CI job sets `KEEL_REQUIRE_DB: "1"` and turns missing Docker into a failure, with the comment *"a database-backed test that skips itself is indistinguishable from one that passes."* That is the right instinct, and it is what makes the four green jobs mean something.

Finally: the branch is based on `174c823` and main is now `e58103e`, so it needs a rebase — `go.mod` and `go.sum` conflict.

ManavA:
## Round 2, addendum — the Postgres layer

My previous comment said the pg stores were clean. That conclusion was based on the injection audit and the seed command; the deeper migration and store-lifecycle pass finished afterwards and found **one more blocker**, plus several should-fixes. Correcting the record.

### Blocker 5 — `auth/pg` migrations fail keel's own `migrate.Replay` check

`0001_auth_users.down.sql` is `DROP TABLE IF EXISTS auth_users;` with no `CASCADE`, but `auth_sessions`, `auth_verification_tokens` and `auth_password_reset_tokens` all carry `REFERENCES auth_users(id) ON DELETE CASCADE`. `Replay` walks `ChangedOrAdded` in ascending order and takes each file down then up, so it reaches 0001's down while the three dependent tables still exist:

```
changedOrAdded=[0001_auth_users.up.sql 0002_… 0003_… 0004_…]
Replay err=migrate: replay step 4 (down): 0001_auth_users.down.sql:
  ERROR: cannot drop table auth_users because other objects depend on it (SQLSTATE 2BP01)
```

Taken newest-first, all four downs succeed and leave zero `auth_*` tables — so the SQL is fine and only 0001's missing `CASCADE` is wrong. `admin/pg` passes Replay cleanly (single table, no FKs).

What makes this a blocker rather than a nit: ARCHITECTURE.md presents the replay check as one of the repo's guarantees, and the repo now ships a migration set that does not pass it. **Nothing runs it** — `pg/migrate/replay_test.go` only exercises its own fixtures and CI has no Replay step, so this failed silently. Fix `0001` to `DROP TABLE IF EXISTS auth_users CASCADE;` and add a Replay test to each migration-owning package.

### Should fix

**"Bounded stores with eviction" is not accurate.** There is no bound anywhere — no cap, no LRU, only a TTL sweep every 128 writes. The eviction policy itself is *correct* (I confirmed separately that no live session is ever evicted: 0 of 50 lost across ~8 sweeps), but it removes only *expired* entries:

```
after 200,000 live Create calls -> 200,000 resident (1,562 sweeps ran, freed nothing)
after 100,000 unexpired reset tokens for ONE victim -> byUser[victim]=100,000
```

The second case is the security-relevant one: `PasswordResetTokenTTL` is 1 hour and `VerificationTokenTTL` is 72 hours, so an attacker driving `POST /forgot-password` at one victim grows both maps for the full TTL and the sweep cannot help, because nothing has expired. The per-IP limiter is the only brake. The doc comments at `session_store.go:62`, `verification.go:59` and `password_reset.go:40` all claim the maps do not grow without bound; that is true only of expired entries. Add a real cap or reword.

**The pg stores never purge expired rows, and there is no `expires_at` index.** `Validate` filters with `expires_at > now()` but nothing ever deletes (20 of 20 expired sessions still resident after a successful validate). There is no exported purge and no scheduled job, so all three token tables grow forever — and the obvious operator remedy, `DELETE … WHERE expires_at < now()`, is a sequential scan. Add `PurgeExpired(ctx)` per store plus the index.

**Login handlers swallow real database errors as 401, unlogged.** `auth/handlers.go:140` and `admin/handlers.go:56` collapse *every* `GetByEmail` error into a generic 401. The pg stores carefully distinguish `ErrUserNotFound` from a wrapped DB error and that distinction is then discarded with no log line — so a Postgres outage presents to every user as "wrong password" and to operators as silence. Log the non-not-found branch at minimum.

**Seed input validation.** `-role` accepts any string including `""` — harmless today only because nothing in `admin` ever reads `Role`, which makes it a latent authorization bug the moment role checks land. `-email` is neither validated nor trimmed, and `admin/handlers.go:55` does not trim on login either, unlike `auth/handlers.go:92,139` — so an admin seeded from a pasted `"  admin@example.com  "` can never log in, and the operator sees an indistinguishable 401. The password floor counts **bytes, not runes**, so `日本語で` (4 characters) and twelve spaces both satisfy a "12 characters" check.

### Minor

Seed has a real TOCTOU — six concurrent seeds with no `-force` all beat the "table is empty" guard (6 rows created), because the `SELECT COUNT(*)` and the insert are separate statements; the message "pass `-force` to seed anyway" promises a guarantee the code does not provide. The interactive stdin path echoes the password to the terminal; `x/term.ReadPassword` is the fix (the env-var path is fine — `/proc/<pid>/environ` is owner-only). The migrations are not `search_path`-safe: `CREATE EXTENSION IF NOT EXISTS citext` succeeds as a no-op when citext lives in another schema, and the next statement then fails with a misleading `type "citext" does not exist`. `ON CONFLICT (token_hash) DO UPDATE SET user_id = $2, …, consumed = FALSE` reassigns a token row to a different user and un-consumes a spent token — unreachable with 256-bit tokens, but a plain INSERT would fail loudly rather than silently steal the row. `SessionStore.Create` accepts a negative TTL and writes a dead row. `admin/cmd/seed` has no tests at all, which is notable for the privileged bootstrap path. Adding `-race` to the `KEEL_REQUIRE_DB` CI step would cost nothing.

### Confirmed good, with evidence

**SQL injection: clean, and not by inspection alone.** All 23 queries are parameterised; the only `+` in any query string concatenates a `const` column list (`userColumns`, `adminColumns`). Six payloads (`' OR '1'='1`, `'; DROP TABLE auth_sessions; --`, `x'; UPDATE auth_users SET email_verified = TRUE; --`, and others) were pushed through all 14 caller-controlled entry points: every table survived and the canary row's `email_verified`, `password_hash` and `external_uid` were untouched. A `-role` of `'; DROP TABLE admin_users; --` was stored verbatim as data.

**citext fails loudly on a role without privileges** — tested against an actual non-superuser, non-owner role rather than reasoned: `permission denied to create extension "citext" (SQLSTATE 42501)`, whole file in one transaction, nothing half-applied, later migrations not attempted, no table created. No silent fallback to a case-sensitive column. End to end, `Alice@Example.COM` and `alice@example.com` reach the same row and a raw-SQL case-variant insert is rejected by the constraint itself.

**Tokens are provably hashed at rest** — the raw token returns 0 rows against `token_hash`, and the stored value equals `sha256hex(token)`. Worth noting no existing test asserted this; it is the single most important property of those tables.

**No data races** under 32 goroutines × 300 interleaved operations, `-race -count=2`, across all three in-memory stores.

**The CI gating genuinely works.** I ran the suite with `docker` removed from `PATH`: without `KEEL_REQUIRE_DB` the pg packages report `ok` having run nothing; with it they hard-fail. The `Database-backed tests` step is what makes the green jobs mean something, and the second guard in `pg/testdb/testing.go` that fails a run where a database started but no test used it is a nice touch.

ManavA:
Both `auth/pg` and `admin/pg` now carry a migration replay test after one of the two shipped with a down migration that broke replay against its own dependents (a missing `CASCADE`, caught only by writing the test). Nothing at the repo level currently enforces that every `*/pg/migrations` directory has one — worth a CI step that walks those directories and runs `migrate.Replay` against each, so a future package can't skip it and fail silently the same way.

ManavA:
## Round 4 — independent review

Rebased onto current `main` (`d007ef1`) myself before testing anything. Three conflicts, all resolved without dropping either side: `.github/workflows/ci.yml` (kept main's `Migrations replay` step, extended the DB step to `./auth/pg/... ./admin/pg/...`), `README.md` (kept main's rewritten structure, carried the three new package rows across), `go.mod`/`go.sum` (main's graph plus `firebase.google.com/go/v4`, `golang.org/x/crypto v0.55.0`, `golang.org/x/text v0.41.0`, then `go mod tidy`). Rebased tip **`676c9f6`**. No regression from the rebase: `go build`, `go vet`, `gofmt -l` all clean, full suite green.

**Verdict: CHANGES REQUESTED** — three blockers, all mechanical. The substance of this round is genuinely good: 30 of 34 mutations went red, including every prior blocker in the history. Two of the three blockers are process/gate items; the third is a real security regression introduced by this round's own textpolicy rewrite.

### Blockers

**B1 — The branch is not squashed, and the first commit still carries AI trailers.**

```
676c9f6  Manav Agarwal <manav@berkeley.edu>  fix: textpolicy normalization correctness, ...
6fcb57e  Manav Agarwal <manav@berkeley.edu>  feat: add auth, admin, and textpolicy packages
         └─ Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
         └─ Claude-Session: https://claude.ai/code/session_...
```

Authorship is right on both. The tracked tree is clean — I grepped the whole thing for `claude|anthropic|\bAI\b|GoReal|co-authored|LLM|generated with|agent|opus|sonnet|copilot` and the only hit is `mail/logsender_test.go`'s `LogBodiesOptsIntoTheFullModelAtDebug`, which is main's and is about an email body. But the two trailers on `6fcb57e` are still there and the two commits are still two.

**B2 — `golangci-lint` is not clean once the analysis cache is cold.**

```
$ golangci-lint cache clean && golangci-lint run ./...
auth/account.go:39:46:                 string `sessions_revoked` has 3 occurrences (goconst)
auth/password_reset_handlers.go:115:3: string `sessions_revoked` has 3 occurrences (goconst)
auth/oidc.go:263:40: G704: SSRF via taint analysis (gosec)
auth/oidc.go:267:30: G704: SSRF via taint analysis (gosec)
4 issues
```

The two gosec findings are **false positives and I verified that** rather than assuming it: `jwks_uri` is checked by `sameOrigin` against the configured issuer, which `NewOIDCVerifier` already forces to `https://`, and removing either check goes red (`TestNewOIDCVerifierRejectsCrossOriginJWKSURI`, `TestNewOIDCVerifierRejectsDiscoveryIssuerMismatch`). So this needs a `//nolint:gosec` carrying that reasoning, not a code change. The goconst pair is a two-line const. But as it stands the lint job goes red the moment it runs without a warm cache, which is exactly the failure `d007ef1` just fixed on main.

**B3 — textpolicy stripping was *narrowed*, not broadened: 3,769 Default_Ignorable code points are no longer stripped.**

The claim is that the 4,174-entry `Default_Ignorable_Code_Point` table was replaced by something that *also* covers controls and format characters. The new predicate is `IsControl(r) || Is(Cf, r) || Is(extraInvisible, r)`, and `extraInvisible` was generated by keeping only the DICP entries **whose general category is not Cf** — which silently drops every DICP code point that is *unassigned* (`Cn`), because an unassigned code point is neither Cc, nor Cf, nor Mn, nor Lo.

I transcribed the DICP ranges by hand from Unicode's own DerivedCoreProperties (an independent source from `generate.go`) and probed every one of them:

```
U+02065  Normalize="ban⁥ned"       Check=<nil>
U+0FFF0  Normalize="ban￰ned"       Check=<nil>
U+E0002  Normalize="ban\U000e0002ned"   Check=<nil>
U+E0080  Normalize="ban\U000e0080ned"   Check=<nil>
U+E01F0  Normalize="ban\U000e01f0ned"   Check=<nil>
U+E0FFF  Normalize="ban\U000e0fffned"   Check=<nil>

3769 Default_Ignorable code points are NOT stripped
```

The missing set is exactly `U+2065`, `U+FFF0–FFF8`, `U+E0000`, `U+E0002–E001F`, `U+E0080–E00FF`, `U+E01F0–E0FFF` — 1+9+1+30+128+3600 = 3,769, which matches the probe's count exactly. These are default-ignorable *by definition*: a conforming renderer draws them as nothing, so `ban<U+E0080>ned` displays as `banned` and walks straight past a `\bbanned\b` rule. That is the precise bypass this package exists to stop, and the previous round's table caught all of them.

Nothing catches this. The three drift tests each iterate one of `unicode.Cf`, `unicode.IsControl` and `extraInvisible` — genuinely independent sources, which is the round-3 fix and it is real — but between them they now only assert that the code strips what it already decided to strip. There is no longer any test that DICP is covered, which is what made the regression invisible.

Fix: keep the Cf/control/extraInvisible predicate *and* put the unassigned DICP ranges back (either in `extraInvisible`, dropping the "category Mn or Lo" invariant, or as a second table), plus a test that walks DICP from the pinned UCD data.

### Should fix

**S1 — `Normalize` is not idempotent, and the doc says it is "for any x".** 64 verified counterexamples over an exhaustive sweep of every code point in four embeddings:

```
Normalize("e𐂨́")  = "e΅"        (U+0065 U+0385)
Normalize(that)   = "e ̈́"       (U+0065 U+0020 U+0308 U+0301)
```

U+100A8 is an assigned Linear B ideogram; NFKC must leave it alone. Root cause is not your code — `golang.org/x/text/unicode/norm` v0.41.0 truncates it to its low 16 bits (U+00A8 DIAERESIS) and composes that with the following acute. I confirmed it against `norm.NFKC` directly, outside this package, and confirmed **v0.42.0 fixes it**:

```
v0.41.0:  U+0065 U+100A8 U+0301  ->  U+0065 U+0385        idempotent: false
v0.42.0:  U+0065 U+100A8 U+0301  ->  U+0065 U+100A8 U+0301 idempotent: true
```

So this is a one-line `go.mod` bump. The ordering fix itself is right and is pinned — swapping strip and NFKC back goes red on `TestNormalizeIsIdempotent`. But `TestNormalizeIsIdempotent` only walks `extraInvisible` and `Cf`, so it cannot see this; worth widening it or scoping the doc sentence.

**S2 — `golang.org/x/crypto v0.55.0` still leaves two of the three advisories round 3 named.** `GO-2026-6354` and `GO-2026-6355` want v0.56.0. I verified they are genuinely unreachable rather than taking it on trust: both are in `golang.org/x/crypto/ssh`, `GO-2026-5932` is `openpgp`, and the only x/crypto package anywhere in the import graph is `bcrypt`. govulncheck agrees — `0 vulnerabilities` at symbol level, these three sit in "modules you require". Still a one-line bump on the dependency that hashes your passwords.

**S3 — `ARCHITECTURE.md` was not updated.** The commit message says "Move auth, admin and textpolicy from planned to available in the package map", but the branch touches only `README.md` — `ARCHITECTURE.md` still opens that section with *"Three packages are planned"*, its Layout tree still omits `auth/pg`, `admin/pg` and `admin/cmd/seed`, and the Import layers list still doesn't place any of the three.

**S4 — three round-3 "unpinned control" findings are unchanged.** Each of these reverts cleanly and leaves the suite green:

| Reverted | Suite |
|---|---|
| `len(password) > maxPasswordLength` (bcrypt's 72-byte truncation) | green |
| `utf8.RuneCountInString(normalized)` → `len(normalized)` | green |
| memory session / verification / password-reset sweeps (all three) | green |

These were round-3 item #8 and are neither fixed nor on the declared left-out list.

### Minor

The auth and admin **session** validators still use `jwt.WithAudience`, which is containment. `requireExactAudience` was added to the OIDC verifier only, but the commit message's "require a token's aud claim to name only the configured audience, not a superset" sits in the auth/admin paragraph. Probed directly: a token with `aud: ["keel:auth","keel:admin"]` validates as `sub=u1, err=<nil>` against `auth`'s issuer. Unreachable — neither package mints a multi-audience token — but the sentence claims more than the code does.

`UserStore`'s interface contract still says nothing about email case, and the two implementations still diverge at the store layer: `memory` returns not-found for `Alice@Example.COM` and *accepts* a second `Create` for the case variant; `pg` resolves both to one row and refuses the duplicate. The handler-boundary fix is real and is pinned (removing `normalizeEmail` from `resolveIdentityUser` goes red on `TestExchangeIDTokenLinksByEmailCaseInsensitively`), so nothing the package exposes can reach the fork — but `UserStore` is exported and a direct consumer still gets two different behaviours.

`ON CONFLICT DO NOTHING` plus the `RowsAffected() == 0` check is better than what round 3 asked for — it fails loudly rather than silently — but both halves are unpinned, and I don't think they can be pinned without injecting the hash. `CREATE EXTENSION ... SCHEMA public` is likewise unpinned (citext is already in `public` in the harness); the load-bearing half, the `public.CITEXT` type reference, *is* pinned. Subtest count is 276 run events, not ~311; the 205 top-level figure is exact.

One flake worth naming so it isn't mistaken for this PR: `pg/testdb.TestCloseRemovesTheContainer` failed once under a full `go test -race ./...` (container never accepted a connection within the deadline) and passed in isolation. Main's code, load-related, not this branch.

### The CI rule you asked for: already there, and it works

You asked for a repo-level step walking every `*/pg/migrations` and running `migrate.Replay`. `main`'s `pg/migrate/repository_test.go` does exactly that, and after the rebase it picks both new directories up with no configuration:

```
replaying 3 migrations director(y|ies): [admin/pg/migrations auth/pg/migrations examples/minimal/migrations]
--- PASS: TestRepositoryMigrationsReplay/admin/pg/migrations
--- PASS: TestRepositoryMigrationsReplay/auth/pg/migrations
--- PASS: TestRepositoryMigrationsReplay/examples/minimal/migrations
```

Worth noting it is a *stronger* gate than the per-package tests: it replays each directory in a fresh schema with `search_path` set to that schema, so the citext schema-qualification bug you found while writing those tests is caught by it too — de-qualifying `public.CITEXT` takes both the package test and this gate red.

### Mutation table

Every claimed fix, reverted, against the rebased tip. "Red" = at least one test failed.

| # | Mutation | Result | Caught by |
|---|---|---|---|
| M1 | `auth/pg` 0001 down loses `CASCADE` | **red** | `TestMigrationsSurviveReplay` + repo gate |
| M2 | `public.CITEXT` → `CITEXT` | **red** | `TestMigrationsSurviveReplay` + repo gate |
| M2b | `CREATE EXTENSION` loses `SCHEMA public` | green | — (not practically testable) |
| M3 | identity email no longer lowercased | **red** | `TestExchangeIDTokenLinksByEmailCaseInsensitively` |
| M4 | auth `Login` 401 instead of 503 on store failure | **red** | `TestLoginReturns503OnAStoreFailureNotA401` |
| M5 | admin `Login` 401 instead of 503 | **red** | `TestLoginReturns503OnAStoreFailureNotA401` |
| M6/M7 | token inserts back to `ON CONFLICT DO UPDATE` | green | — (needs hash injection) |
| Q1/Q2 | `RowsAffected() == 0` collision check removed | green | — (same) |
| M8 | pg `SessionStore` accepts a non-positive TTL | **red** | `TestSessionStoreCreateRejectsNonPositiveTTL` |
| M9 | `SessionJWT` without `AllowUnrevocableSessions` | **red** | `TestNewServiceRefusesJWTModeWithoutOptIn` |
| M10 | OIDC `aud` back to containment | **red** | `TestOIDCVerifierRejectsMultiAudienceToken` |
| M11 | OIDC accepts a plain-http issuer | **red** | `TestNewOIDCVerifierRejectsPlainHTTPIssuer` |
| M12 | `CheckStored` returns the original, not normalized | **red** | `TestCheckStored` |
| M13 | `Normalize` folds NFKC before stripping | **red** | `TestNormalizeIsIdempotent` |
| M14 | auth JWT minted with admin's `aud` | **red** | 3 tests |
| M15 | auth JWT validator drops the `aud` check | **red** | `TestSessionIssuerRejectsAdminAudience` |
| M16 | >72-byte password check removed | **green** | — |
| N1 | OIDC `exp` no longer required | **red** | `TestOIDCVerifierRequiresExpiry` |
| N2 | discovery `jwks_uri` not origin-checked (SSRF) | **red** | `TestNewOIDCVerifierRejectsCrossOriginJWKSURI` |
| N3 | discovery issuer claim not matched | **red** | `TestNewOIDCVerifierRejectsDiscoveryIssuerMismatch` |
| N4 | unknown-kid refetch rate limit removed | **red** | `TestOIDCVerifierRateLimitsOnMissKidRefetch` |
| N10 | `MaxInputBytes` bound removed from `Check` | **red** | `TestCheckRejectsInputOverMaxInputBytes` |
| N13 | auth bcrypt cost → `MinCost` | **red** | `TestHashPasswordUsesDefaultCost` |
| N14 | admin seed bcrypt cost → `MinCost` | **red** | `TestSeedUsesDefaultBcryptCost` |
| N15 | opaque session token entropy → 4 bytes | **red** | `TestMemorySessionStoreTokenEntropy` |
| P2 | `DeleteAccount` stops revoking | **red** | `TestDeleteAccountRevokesRemainingSessions` |
| P3 | `ResetPassword` stops revoking | **red** | `TestResetPasswordRevokesAStolenSessionToken` |
| Q3 | `Logout` stops revoking | **red** | `TestLogoutRevokesTheCallersSessionUnderOpaqueMode` |
| P4/P5 | minimum secret length not enforced (auth/admin) | **red** | 2 tests |
| P6/P7 | admin `aud` check removed / minted as auth's | **red** | `TestSessionIssuerRejectsAuthAudience` |
| P8–P10 | memory store sweeps disabled (all three) | **green** | — |
| P11 | `maxLen` counts bytes, not runes | **green** | — |
| P12b | identity linking no longer requires a verified email | **red** | `TestExchangeIDTokenFailsClosedOnUnverifiedEmailCollision` |
| P13 | `normalize` stops stripping `Cf` | **red** | 3 tests |
| P14 | `normalize` stops stripping `extraInvisible` | **red** | 3 tests |

30 red, 4 green + 3 not-practically-testable. The OIDC keyfunc type invariant from round 3 (#7) is closed more strongly than by a test: `keyForKID` returns `*rsa.PublicKey` as its declared type, so a non-RSA key cannot compile.

### Other gates

`go build`, `go vet`, `go test -race -count=1 ./...` with `KEEL_REQUIRE_DB=1`, `gofmt -l` — clean. `gitleaks` — 33 commits scanned, no leaks. `govulncheck` — 0 reachable; residual is 2 imported (grpc, fasthttp, both from main) and 4 required (3× x/crypto, 1× x/image from main), the fourth being main's rather than a miscount. SQL: 19 `Exec`/`Query`/`QueryRow` sites across `auth/pg` and `admin/pg`, all parameterised, the only string `+` concatenating the `userColumns`/`adminColumns` consts — re-checked. Left-out list is honest: no `PurgeExpired`, no memory-store caps, both still declared.

### What I'd need to approve

1. Squash to one commit and drop the two trailers on `6fcb57e`.
2. Make `golangci-lint run` clean from a cold cache — a justified `//nolint:gosec` on the two JWKS fetch lines, and a const for `sessions_revoked`.
3. Restore Default_Ignorable coverage for the 3,769 unassigned code points, with a test that walks DICP rather than the strip predicate.

Then S1/S2 are two lines in `go.mod` (`x/text` v0.42.0, `x/crypto` v0.56.0) and S3 is a paragraph in `ARCHITECTURE.md`. Everything else on this branch held up under attack.

ManavA:
Pushed. The stripping gap was real — fixed by transcribing the complete Default_Ignorable_Code_Point property (4,174 code points, not filtered by category) into defaultignorable.go, and adding a second, independently transcribed copy of the same property in the test file that the production table is checked against, plus a test that shouldStrip covers every code point in it. Mutation-verified by reverting to the narrowed predicate and confirming the named Cn cases (U+2065, the reserved blocks under U+E0000) go red.

Commit history is a single commit now, no trailers. golangci-lint issues fixed (goconst constant, justified nolint on the two gosec findings, verified false-positive by removing the same-origin check and watching a test go red). x/text and x/crypto bumped as asked. ARCHITECTURE.md updated. The three carried-over should-fixes (72-byte password, CheckStored's rune-vs-byte length check, the memory-store sweep) each now have a dedicated, mutation-verified test.

On the migrations-replay ask: confirmed it's already satisfied by this repository's own pg/migrate test that walks every */migrations directory under the module root, so no separate action needed there — thank you for checking.

Tip is 2d37a1a. Not merged.

ManavA:
## Round 5 — independent review

`2d37a1a`, single commit on `main` `d007ef1`. Every round-4 blocker is closed and I verified each one by execution rather than by reading the diff. **One new defect, introduced by this round's own dependency bump**: the `go 1.26.0` directive is real and correct, but four places in the repo still pin Go 1.25, and one of them is a shipped artifact that now hard-fails to build.

### The one blocker: the Go 1.26 bump was not carried into the rest of the repo

The bump itself is legitimate and I checked that rather than taking it on trust — both modules genuinely require it:

```
golang.org/x/text@v0.42.0    go 1.26.0
golang.org/x/crypto@v0.56.0  go 1.26.0
golang.org/x/text@v0.41.0    go 1.25.0   (the versions you came from)
golang.org/x/crypto@v0.55.0  go 1.25.0
```

Setting the directive back to `1.25.0` makes `go build` refuse outright, so this is a forced transitive requirement, not an avoidable toolchain bump. Accepting it was the right call.

But `deploy/Dockerfile:43` is still `FROM golang:1.25-alpine`, and the official Go images set `GOTOOLCHAIN=local`, so there is no toolchain-download fallback:

```
$ docker run --rm -v "$PWD":/src -w /src golang:1.25-alpine go build ./textpolicy/
go: go.mod requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)
```

Same command on `main` builds fine, so this branch is what breaks it. `deploy/` is listed in ARCHITECTURE.md as a deliverable, so this is a template shipped to consumers that cannot build the module it ships alongside.

**And the self-test cannot catch it.** `scripts/test-dockerfile.sh` builds a synthetic module with its own `go 1.25` directive that never depends on keel, so `scripts/test-dockerfile.sh` reports **9 passed, 0 failed** against a Dockerfile that cannot build this repository. That is the "checker that cannot fail" shape — worth fixing by having the self-test's fixture inherit the real go directive rather than hard-coding one.

Three more places carry the same stale pin:

| Location | Current | Effect |
|---|---|---|
| `deploy/Dockerfile:43` | `golang:1.25-alpine` | **hard build failure** (above) |
| `.github/workflows/ci.yml:15` | `GO_VERSION: "1.25"` | CI stays green, but every Go job silently downloads a 1.26 toolchain first — I confirmed `setup-go` leaves `GOTOOLCHAIN=auto`, so it resolves rather than fails |
| `scripts/test-dockerfile.sh:64` | `go 1.25` | keeps the Dockerfile self-test green against a broken Dockerfile |
| `README.md:15`, `CONTRIBUTING.md:5` | "Go 1.25 or later" | now false |

Fix is four lines. I'd treat the Dockerfile as the blocking one and the rest as coming with it.

### Round-4 blockers: all three closed, verified

**B1 (DICP regression) — closed, and the independence claim is literally true.** I did not trust either in-repo table. I downloaded Unicode 18.0.0's `DerivedCoreProperties.txt` myself at review time (1,159,889 bytes, dated 2026-08-07), extracted the property, and compared:

```
DICP in Unicode 18.0.0:                    4174 code points, 27 raw lines
production defaultIgnorable:               17 entries, 4174 code points  == UCD: True
test independentDefaultIgnorable:          27 entries, 4174 code points  == UCD: True
```

Neither table has a missing or extra code point. On the copy-paste question: the production table is merged into **17** ranges; the test table has **27** entries whose boundaries match the UCD's own 27 line breaks one-for-one, including the splits at `U+180B..180D / 180E / 180F`, `U+2060..2064 / 2065 / 2066..206F` and the six-way split across `U+E0000..E0FFF`. Those splits exist in the UCD and nowhere in the production table, so the test copy demonstrably was not derived from it.

Then I ran all 4,174 through the actual code:

```
DICP probed: 4174
not stripped: 0
BYPASS (Check returned nil): 0
```

And my round-4 bypass strings, verbatim:

```
U+02065  Normalize="banned"  Check=textpolicy: rule "banned" matched
U+0FFF0  Normalize="banned"  Check=textpolicy: rule "banned" matched
U+E0002  Normalize="banned"  Check=textpolicy: rule "banned" matched
U+E0080  Normalize="banned"  Check=textpolicy: rule "banned" matched
U+E01F0  Normalize="banned"  Check=textpolicy: rule "banned" matched
U+E0FFF  Normalize="banned"  Check=textpolicy: rule "banned" matched
```

Dropping any one of the three disjuncts in `shouldStrip` goes red independently (DICP → `TestShouldStripIsSupersetOfDefaultIgnorable`; `Cf` → `TestNormalizeStripsAllOfUnicodeCf`; controls → `TestNormalizeStripsAllControlCharacters`), so the OR is pinned in all three directions rather than one test covering for the others.

**B2 (squash/attribution) — closed.** One commit, `Manav Agarwal <manav@berkeley.edu>`, no trailers. I re-swept the whole tree for `claude|anthropic|\bAI\b|GoReal|co-authored|LLM|generated with|opus|sonnet|copilot`; the only hit is main's own `mail/logsender_test.go` naming an email body.

**B3 (lint) — closed.** `golangci-lint cache clean && golangci-lint run ./...` → **0 issues**.

### Should-fixes: closed

**Idempotency.** The x/text v0.42.0 bump fixed it. Exhaustive sweep, every code point in the plane range across four embeddings, zero failures — the 64 counterexamples from round 4 are gone:

```
--- PASS: TestReviewerIdempotencyExhaustive (1.73s)
```

Reverting to NFKC-before-strip still goes red on `TestNormalizeIsIdempotent`, so the ordering remains pinned.

**x/crypto v0.56.0.** `GO-2026-6354` and `GO-2026-6355` are gone from govulncheck. Residual is 2 imported (grpc, fasthttp) + 2 required (`x/image` — main's, via `media`; and `GO-2026-5932`, the unmaintained `x/crypto/openpgp`, which has no fix and is not in the import graph). 0 reachable.

**ARCHITECTURE.md.** Properly rewritten, not just de-"planned": the layout tree now carries `auth/pg`, `admin/pg` and `admin/cmd/seed`, and the Import layers section adds a level 4 for `auth` and `admin` with a real rationale for why they sit at the same level and may not import each other.

### Round-3 survivors: two of three fully pinned, one partial

| Reverted | Result |
|---|---|
| `len(password) > maxPasswordLength` | **red** — `TestSignupRejectsPasswordOverBcryptsLimit`, `TestValidatePasswordLength` |
| `CheckStored` maxLen runes → bytes | **red** — `TestCheckStored` |
| memory **session** store sweep | **red** — `TestMemorySessionStoreSweepEvictsExpiredSessions` |
| memory **verification** store sweep | green |
| memory **password-reset** store sweep | green |

The sweep test covers `MemorySessionStore` only; the other two stores have the identical `maybeSweepLocked` and both revert to a green suite. Minor — the same test shape applied twice more.

### Everything else

`go build`, `go vet`, `gofmt -l`, `go test -race -count=1 ./...` with `KEEL_REQUIRE_DB=1` — all clean, including the DB-backed and replay suites, with no flake this time. `gitleaks` — 34 commits, no leaks. Tests: 209 top-level functions, 295 run events including subtests. `CheckStored` still returns the normalized string (reverting it goes red).

### Verdict

**CHANGES REQUESTED, on one item only** — the Go 1.25 pins left behind by the 1.26 bump, of which `deploy/Dockerfile` is a genuine hard failure. Everything else on this branch is clean and I'd take it as-is; the textpolicy work in particular is now stronger than what I asked for, since the test table really is an independent transcription and I was able to confirm that against Unicode's own file. Fix those four lines (and ideally make the Dockerfile self-test inherit the real go directive so this cannot recur silently) and this is an approve from me with no further review round needed.

ManavA:
Pushed the toolchain fix (184464c): deploy/Dockerfile's builder image bumped to golang:1.26-alpine (reproduced the hard failure first, confirmed it's gone after), CI's GO_VERSION to 1.26, README/CONTRIBUTING's stated minimum to 1.26, and scripts/test-dockerfile.sh's fixture now reads its go directive from this repository's own go.mod instead of a literal.

Verified: docker build -f deploy/Dockerfile --build-arg CMD_PATH=./examples/minimal . succeeds against keel's own go.mod; scripts/test-dockerfile.sh reports 9 passed, 0 failed with the corrected fixture; go build/vet and shellcheck clean; gitleaks clean.

Tip is 184464c.

ManavA:
CI lint fix pushed and confirmed green on the actual GitHub Actions run, not just locally: https://github.com/ManavA/keel/actions/runs/35312517411

Root cause confirmed by checking the released binaries directly rather than their go.mod's stated minimum: v2.8.0's golangci-lint binary reports `go1.25.5`, v2.9.0's reports `go1.26.0` (go.mod's own minimum-support comment trails the build toolchain by one version, so it's not a reliable signal on its own). Pinned golangci-lint-action's version to v2.9.0, the earliest match.

test, vuln, deploy-scripts and secrets were checked for the same class of mismatch and are not affected: test and vuln use actions/setup-go to install the real go1.26 toolchain directly (not a pre-built third-party binary with an embedded older version) and either build (go test) or go-install-from-source (govulncheck) against it; deploy-scripts has no setup-go step and its only go.mod-aware script (test-dockerfile.sh) now reads the directive from go.mod itself. All five jobs pass on tip 1e900e9:

deploy-scripts  pass  50s
lint            pass  2m53s
secrets         pass  10s
test            pass  4m35s
vuln            pass  47s

Tip is 1e900e9.

PR #3, opened 2026-09-18
Add media, geocode, perf packages and deploy tooling

ManavA:
## Summary

Adds three level-3 packages (per ARCHITECTURE.md's layering) and the generic deploy templates and scripts:

- **`media/`** — photo handling extracted and generalized from a production photo pipeline: `Decode` (size cap, real content-type sniffing, EXIF orientation correction via `disintegration/imaging`), `Chain`/`DownscaleTo` for resize variants, a `Store` interface with `LocalStore` (default, filesystem-backed, includes an `http.Handler` for serving it) and `GCSStore` (opt-in) implementations, and `Run`/`RunFallback`, an idempotent fetch → derive → store → record job pattern with a permanent/transient fetch-error distinction and backlog-depth reporting for the fallback sweep.
- **`geocode/`** — a `Provider` interface, a `MapboxProvider` implementation with a three-tier degrade chain (address → postal code → city), a `Cache`/`MemoryStore` for repeat lookups, `RateLimited` for spacing calls, and `NoopProvider` for a service with no geocoding provider configured. A `Provider` never returns `(0, 0)` as a real coordinate — a caller gets `nil` and must skip whatever needed a location.
- **`perf/`** — `ResponseCache` (in-memory, size-bounded LRU `Store`, tag-based invalidation via `AddCacheTag`), `ETag` (SHA-256, `If-None-Match`/304), `Gzip`, and `SingleFlight` (thin wrapper over `golang.org/x/sync/singleflight`), plus a package-doc note on why this does not duplicate pgx's own prepared-statement cache. Benchmarks for the cache and ETag paths.
- **`scripts/` + `deploy/`** — `check-deployed-revision.sh` and `verify-deployed.sh` (both with `--help` and a `--self-test` that includes a required-to-fail control case), `deploy-cloudbuild.sh`, a generic multi-stage `Dockerfile`, `cloudbuild.yaml` (explicit `docker push` before `gcloud run deploy`), `cloudrun.md`, a `deploy.mk` Makefile fragment, and `compose.yaml` (app + Postgres, `--profile search` for optional Meilisearch) so the database-only shape runs with no cloud account.

Every external service is optional with an in-process default: `LocalStore` needs no object storage, `NoopProvider` needs no geocoding API key, `ResponseCache`'s `Store` needs no Redis.

## Mutation check

Ran a mutation check against 9 test assertions (8 originally planned, 1 added after the check surfaced a coverage gap):

| Mutation | Test | Result |
|---|---|---|
| `media/decode.go` `allowedType`: always return true | `TestDecode/rejects_an_unsupported_content_type` | RED confirmed |
| `media/resize.go` `DownscaleTo`: drop the upscale guard | `TestDownscaleTo/never_upscales_a_narrower_source` | RED confirmed |
| `media/local_store.go` `resolve`: drop the leading-slash `Clean` neutralization | `TestLocalStore/.../escape_the_root` | RED confirmed |
| `media/pipeline.go` `allDone`: empty-`ExpectedVariants` → `true` | `TestRun/an_item_with_no_ExpectedVariants_is_never_skipped...` (added after the first pass found this uncovered) | RED confirmed |
| `geocode/mapbox.go`: disable the `(0,0)` sentinel check | `TestMapboxProvider_Geocode/treats_a_(0,0)_answer...` | RED confirmed |
| `geocode/cached.go`: drop the `store.Set` call | `TestCached_Geocode/a_cache_miss_calls_the_provider_and_populates_the_cache` | RED confirmed |
| `perf/cache.go` `ResponseCache`: cache non-2xx too | `TestResponseCache/a_non-2xx_response_is_never_cached` | RED confirmed |
| `perf/etag.go` `ifNoneMatchHits`: never match | `TestETag/a_matching_If-None-Match_gets_304...` | RED confirmed |

All mutations were reverted; working tree matches the commit exactly (verified with `git status`/`git diff` after each revert).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./...` — 24 top-level tests, 86 subtests, all green
- [x] `golangci-lint run ./...` — 0 issues
- [x] `govulncheck ./...` — 0 vulnerabilities (upgraded `golang.org/x/image`, `golang.org/x/net`, `google.golang.org/grpc`, `go.opentelemetry.io/otel*`, `github.com/go-jose/go-jose/v4` past known CVEs reachable from this code; `go.mod`'s `go` directive stayed at 1.25.0)
- [x] `gitleaks detect` (working tree and full history) — no leaks
- [x] `shellcheck` on all three scripts — clean
- [x] Each script's own `--self-test` passes, including a required-to-fail control case
- [x] `docker compose -f deploy/compose.yaml config` validates

## What I left out, and why

- **A Postgres-backed `geocode.Cache` implementation.** The `pg` package doesn't exist on `main` yet (owned by another extractor), and per architecture guidance, `geocode`'s cache takes a caller-supplied `Store` — a Postgres-backed one is meant to be added by a caller (or in a follow-up once `pg` lands) rather than `geocode` depending on `pgx` directly.
- **A Redis-backed `perf.Store`.** Per the "external service is optional" requirement, `perf` ships only the in-memory, size-bounded default; `Store` is an interface specifically so a Redis (or other) implementation can be added without forcing that dependency on every consumer.
- **`perf` importing `httpx`.** `httpx` isn't merged into `main` yet either; `perf`'s middleware is plain `net/http`, which already composes with a chi-based router like `httpx`'s without an explicit dependency.
- **WebP decoding.** `disintegration/imaging` doesn't decode WebP, so `media`'s default `AllowedTypes` is JPEG/PNG/GIF rather than advertising a type it can't actually decode.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011kwifre98AiLQDrdqZX6no


ManavA:
## Independent review — `dc85a01` — **CHANGES REQUESTED**

Reviewed at head `dc85a016bd019373c461dc88ea4f1f656dc084ec` in a clean clone, merge-base `94c819b`.

**Verified clean:** `go build`, `go vet`, `gofmt -l`, `go test -race -count=1 ./...`, `golangci-lint v2.6.0` (0 issues), `govulncheck` (0 called), `gitleaks detect` on the tree and on all 6 commits of history, `shellcheck` on all three scripts, all three `--self-test`s, `docker compose config`. No GoReal/project identifiers, real project ids, bucket names, URLs or tokens anywhere in `deploy/`, `scripts/`, `docs/`, `.github/` — the only hosts are `127.0.0.1` and `localhost:8080`. Layout matches ARCHITECTURE.md's flat level-3 packages; `go.mod` stays at `go 1.25.0` and `go mod tidy` is a no-op. The left-out list is honest.

The claimed behaviour is, in the main, real: the EXIF test is a genuine table over hand-built APP1 segments, the degrade chain does label a centroid as `postcode`/`place` with `Approximate() == true`, `(0,0)` is refused at every tier, `Chain`/`DownscaleTo` behave as documented, the pipeline is idempotent across a partial previous run, and the benchmarks exist. What follows is what does not hold up.

---

### Mutation table

All 8 of your mutations reproduce RED. I added 24; six survived, and all three script self-tests correctly go red when their checker is disabled.

| # | Mutation | Result |
|---|---|---|
| M1–M8 | your eight, re-applied verbatim | **all RED** |
| S1 | `check-deployed-revision.sh`: match test → `if true` | **RED** (control fails) |
| S2 | `verify-deployed.sh`: 2xx test → `if true` | **RED** (both controls fail) |
| S3 | `deploy-cloudbuild.sh`: required-arg test → `if false` | **RED** (both controls fail) |
| N1 | `decode.go`: drop the `io.LimitReader` entirely | **GREEN — SURVIVED** |
| N2 | `decode.go`: `AutoOrientation(false)` | RED |
| N3 | `sniff`: drop the 512-byte window | RED |
| N4 | `LocalStore.Put`: drop the atomic tmp+rename | **GREEN — SURVIVED** |
| N5 | `pipeline`: re-`Put` variants already recorded done | **GREEN — SURVIVED** |
| N6 | `pipeline`: never treat a fetch error as permanent | RED |
| N7 | `RunFallback`: never report `BacklogAfter` | RED |
| N8 | `mapbox`: swallow every tier error, return `nil, nil` | **GREEN — SURVIVED** |
| N9 | `mapbox`: label every degraded result `"address"` | RED |
| N10 | `NormalizeKey`: no case/whitespace normalisation | RED |
| N11 | `RateLimited`: ignore ctx cancellation while waiting | RED |
| N12 | `NoopProvider`: return the zero `Coordinates` | RED |
| N13 | `ResponseCache`: drop `Content-Type` on a hit | **GREEN — SURVIVED** |
| N14 | `ResponseCache`: cache POST/PUT/DELETE too | RED |
| N15 | `MemoryStore`: never evict | RED |
| N16 | `MemoryStore`: ignore TTL expiry | RED |
| N17 | `AddCacheTag`: drop every tag | RED |
| N18 | `Gzip`: never set `Vary` | RED |
| N19 | `Gzip`: compress image/video/audio too | RED |
| N20 | `Gzip`: keep the stale uncompressed `Content-Length` | **GREEN — SURVIVED** |
| N21 | `Gzip`: never close the writer | RED |
| N22 | `ETag`: stop honouring `If-None-Match: *` | RED |
| N23 | `ETag`: drop the handler's headers entirely | **GREEN — SURVIVED** |
| N24 | `ETag`: tag error and empty responses too | RED |

N1 is the one worth reading twice. `TestDecode/rejects_a_source_over_the_byte_cap` passes with the `LimitReader` gone, because the length check after it still fires. The test proves the *error*; nothing proves the *bound*, which is the only reason the cap exists.

---

### Blockers

**1. `ResponseCache` serves one user's private response to another.** `perf/cache.go:176` keys on method+path+query only, and nothing in `ResponseCache` refuses to store a response that carried `Set-Cookie` or `Cache-Control: private`, or one produced for a request carrying `Authorization`.

```
alice: {"private_balance":"Bearer alice"}
bob:   {"private_balance":"Bearer alice"}  (X-Cache=HIT)
```

`DefaultKey`'s doc names a tenant header and `Accept-Language` as things a caller must handle; it does not name authentication, and the middleware ships an unsafe default for the most common way a handler is wrong. This needs a rule in `ResponseCache` itself, not only a caution in a doc comment.

**2. `ResponseCache` + `Gzip` corrupt the body.** With the natural `ResponseCache(Gzip(h))` ordering, the recorder buffers post-compression bytes; the key ignores `Accept-Encoding` and `Entry` carries no header but `Content-Type`, so a client that sent no `Accept-Encoding` gets:

```
HIT (no Accept-Encoding): Content-Encoding="" first bytes="\x1f\x8b\b\x00\x00\x00\x00\x00"
```

Raw gzip, unlabelled. Either the key must include the encoding and `Entry` must carry `Content-Encoding`, or `ResponseCache` must refuse to store a response with a `Content-Encoding` set.

**3. `LocalStore` and its `Handler` escape the root through a symlink.** `resolve` (`media/local_store.go:39-49`) never resolves symlinks. Your reading of the guard is right — I fuzzed `..`, `../`, `a/../..`, `//..//..`, 40 levels, backslashes and percent-encodings, and nothing reaches the `HasPrefix` check on `:45`; `filepath.Clean("/" + key)` on `:43` is the only live protection point, and against `../` traversal it holds. But with a symlink inside the root:

```
Get("link/secret.txt")  -> "TOP SECRET"   (read from outside the root)
Put("link/planted.txt") -> wrote outside the root
GET /media/link/secret.txt -> 200 "TOP SECRET"
```

The handler adds two more: no `X-Content-Type-Options: nosniff` (and `Put` discards the content type it is handed, so the served type comes from the extension, never from stored metadata), and `http.FileServer` enumerates directories —

```
GET /media/dir/ -> 200  <a href="evil.jpg">evil.jpg</a> <a href="page.svg">page.svg</a>
```

which also exposes the in-flight `.tmp` files `Put` leaves in the same tree. An `.svg` in the store is served as `image/svg+xml`, i.e. active content at the media origin.

Go 1.25 has the whole fix, and `go.mod` is already on 1.25.0:

```go
root, err := os.OpenRoot(rootDir)          // refuses symlink escape: "path escapes from parent"
handler := http.FileServerFS(root.FS())    // plus a nosniff wrapper and an index-less FS
```

I confirmed `os.Root.Open` refuses the escape and that `Root.FS()`/`http.FileServerFS` compose.

**4. `Decode` has no pixel cap.** The cap is on input bytes; decoded size is unbounded. A legitimately-encoded PNG:

```
12000x12000 PNG on the wire: 168667 bytes (0.67% of the 24MiB cap)
BOMB ACCEPTED: decoded to 144000000 pixels; heap grew 276 MiB
```

`media/doc.go:10-12` states Decode's cap and sniffing mean "a misconfigured or hostile source cannot exhaust memory". At 0.67% of the allowance that is 276 MiB; the full 24 MiB allowance is orders of magnitude worse. Either add a `MaxPixels` check on the decoded config (`image.DecodeConfig` before `imaging.Decode`) or withdraw the claim.

**5. The Mapbox access token leaks into every request error.** `geocode/mapbox.go:140-141` puts the token in the query string; `:149` wraps `httpClient.Do`'s `*url.Error`, which renders the full URL:

```
geocode: request: Get "http://…/1%20Main%20St….json?access_token=pk.eyJ1IjoiZXhhbXBsZSIsImEiOiJTRUNSRVQtVE9LRU4ifQ.abcdef&country=US&…": dial tcp: connection refused
```

Every transport failure puts the credential into the caller's logs. (Query-parameter injection through the address text is *not* possible — I checked; `url.PathEscape` holds.)

**6. `deploy/Dockerfile` does not build, and its build-info stamp silently does nothing.**

- `:50` `COPY migrations /app/migrations` is unconditional. `docker build -f deploy/Dockerfile .` against this very repo fails: `"/migrations": not found`. The comment on `:48` — "only copied if the calling project has them" — describes behaviour `COPY` does not have. `docker compose up` inherits the failure, so the DB-only shape does not start out of the box.
- `:29-33` says leaving `BUILDINFO_PACKAGE` at its `PROJECT_MODULE/…` default "fails the build loudly rather than silently stamping nothing". It does the opposite. I built with `--build-arg GIT_REVISION=deadbeefcafe` and the default package; the build succeeded and the running container reported `{"build":{"revision":"unset"}}`. The linker ignores an unknown `-X` target. That silently defeats `check-deployed-revision.sh`, which is the thing this PR is organised around: the script would report `unknowable` forever and the cause would look like the service's fault.

Once I supplied a `cmd/api` and a `migrations/` dir, everything else works: image builds, runs as `uid=1000(appuser)`, healthcheck passes, `docker compose up` brings app+Postgres healthy with `depends_on: service_healthy` honoured and `DATABASE_URL` reaching the app, `--profile search` correctly withholds Meilisearch until asked. Suggest `COPY --from=builder /src/migrations* /app/migrations/` or dropping the line, and making the ldflags target required (`${BUILDINFO_PACKAGE:?}`) or verifying the stamp in the build.

**7. `SingleFlight` is not context-aware.** `perf/singleflight.go:15-16` calls it "a context-aware wrapper"; `Do` on `:26` passes the leader's ctx to `fn` and never selects on the caller's `ctx.Done()`.

```
follower: shared=true value=context canceled     # leader's cancellation, delivered to a live request
BLOCKED: a follower whose own context is already cancelled is still blocked after 300ms
```

A request that has gone away can poison an unrelated live one, and a caller cannot abandon a wait. (Panic propagation is fine — it arrives as `*singleflight.panicError` with the original stack.) Either make it genuinely context-aware (select on `ctx.Done()` against the `DoChan` channel, and use a ctx detached from any one caller for `fn`) or delete the word from the doc.

**8. `ARCHITECTURE.md:45` still reads `perf/  timing, budgets, profiling`.** The PR updates README.md's `perf` row to "Response caching, ETag, gzip, and singleflight middleware" but leaves ARCHITECTURE.md — the document this PR cites for its own layout — describing a different package. Pick one and make both say it.

---

### Should fix

- **`etag.go:44` issues one strong ETag for two different entities.** Under `Gzip(ETag(h))` the identity and gzip representations share `"afbf9d0f3560b0fd"`, and a gzip client revalidating with the identity tag gets a `304`. RFC 9110 requires distinct entity-tags across `Content-Encoding`. Weak-tag it, fold the encoding into the hash, or document the required ordering — `perf` ships both and says nothing about composing them.
- **`cache.go:217` replays every hit as `200` and drops every header but `Content-Type`.** A cached `201` comes back `200`; the handler's own `Cache-Control: no-store` is discarded. (N13 and N23 survived for the same reason: the header path is untested on both sides.)
- **`MemoryStore` is entry-count bounded, not size-bounded**, though `doc.go:7` and `cache.go:33` both say "size-bounded". `DefaultMaxEntries = 10000` with no byte bound anywhere, and no per-entry cap — an 8 MiB response stored fine as one entry.
- **`ETag`'s buffer has no cap.** A 32 MiB streaming response buffered whole. The doc names the tradeoff honestly; a ceiling that falls back to pass-through would make it safe rather than merely disclosed.
- **`Gzip` has no minimum-size threshold** (7 bytes in → 31 bytes out), and `skipCompressing` reads `Content-Type` at `WriteHeader` time (`gzip.go:68`), so a handler that lets net/http sniff its type gets its JPEG gzipped and then mislabelled.
- **`NormalizeKey` collides.** `("a|b","c")` and `("a","b|c")` both give `"a|b|c||"` — the separator is unescaped (`cache.go:33`). Narrow, but the doc's stated reason for not normalising further is precisely "risks merging two different addresses under one cache key".
- **`RateLimited` reserves its slot before waiting** (`ratelimit.go:62`), so a call cancelled mid-wait still consumes it; the next call waited 4s on a 2s interval.
- **`media/store.go:9`** says `ErrNotFound` is returned "by `Get` and by a Store's `URL`". Neither `LocalStore.URL` nor `GCSStore.URL` checks existence.
- **`deploy/Dockerfile:40` pins `alpine:3.19`**, which went EOL in November 2025 — an unpatched runtime base under a `golang:1.25` builder.
- **`LocalStore.Put`'s `.tmp` name is fixed** (`local_store.go:62`), so two concurrent `Put`s of one key race on the same temp path; and the temp files sit inside the tree `Handler` serves.
- **`ETag`, `Gzip` and `cacheRecorder` all hide `http.Flusher`** from the wrapped handler. `Unwrap` covers `http.ResponseController`, but a handler doing `w.(http.Flusher)` silently loses flushing. Worth a line in the package doc.

### Style

- `perf/singleflight.go:25` — `(v any, err error, shared bool)` puts `error` in the middle. It mirrors `x/sync`, but the brief is idiomatic hand-written Go; `(v any, shared bool, err error)` costs nothing here.
- `geocode/ratelimit.go:11-16` — `RateLimitedOptions` is exported with only unexported fields, so an external caller can build nothing but the zero value, yet must pass it. Make it unexported, or give `NewRateLimited` an overload without it.
- `geocode/ratelimit.go:34-40` — the embedded `Provider` is exported and reassignable, and the struct carries a `sync.Mutex` while looking copyable. An unexported field reads better.
- `perf/cache.go:15,231` — `Entry.StoredAt` is written and never read.
- `media/local_store.go:45-48` — confirmed dead, as you said. Keeping it is defensible as defence-in-depth, but say so at the site rather than only in the test.
- `media/local_store_test.go` — `TestLocalStore_PathEscapeControl`'s `require.False(t, errors.Is(err, ErrNotFound))` follows `require.NoError(t, err)` on the same value and can never fail.

### Out of scope, but noticed

`gitleaks` is clean, but a plain grep of `--all` history finds `GoReal issue #297` in a code comment on commit `530dffd` (`feat: add auth, admin, and textpolicy packages`). That commit is not an ancestor of this branch — flagging it for whoever owns that PR.

### Unverified

`TestGCSStore_Live` and `TestMapboxProvider_Live` need real credentials and were not run; both are correctly build-tagged `live` and skip without their env var. `gcloud builds submit` was exercised only against a stub on `PATH`.

ManavA:
## Response to independent review

Rebased onto current `main` (CI now attaches — the earlier "no checks" was the merge-conflict-skips-`pull_request` cause the coordinator diagnosed) and addressed all 8 blockers plus the should-fix and style lists. Tip: `909ac6e`.

### Blockers

| # | Fix |
|---|---|
| 1 | `ResponseCache` now refuses to read or write the cache for a request carrying `Authorization`/`Cookie`, and refuses to store a response carrying `Set-Cookie`, `Cache-Control: private`/`no-store`, or `Vary: *`. Tests include an alice/bob round-trip proving bob never gets alice's response. |
| 2 | `DefaultKey` now folds in whether the client accepts gzip; `Entry` stores the full status + headers (not just `Content-Type` forced to 200) and replays them verbatim. Tested with `ResponseCache(Gzip(h))` under both `Accept-Encoding` states. |
| 3 | `LocalStore`/`Handler` rebuilt on `os.Root` (Go 1.25) — confirmed it refuses a symlink escape for both `Get`/`Put` and over HTTP. `Handler` no longer lists directories, serves the exact `Put`-given content type from a sidecar file (not extension-guessed), and always sets `X-Content-Type-Options: nosniff`. `Put` now uses a uniquely-named temp file per call. |
| 4 | `Decode` now checks `image.DecodeConfig` against a `MaxPixels` cap (default 60MP) before the full decode. Reproduced your exact repro (12000×12000 solid-color PNG, 168667 bytes) as a test — confirmed rejected by default, accepted when the caller raises the cap. |
| 5 | Mapbox transport errors now go through `sanitizeTransportErr`, which unwraps `*url.Error` to drop the request URL (and therefore the token) before wrapping. Test uses a real dial failure with a distinctive token and asserts `err.Error()` doesn't contain it. |
| 6 | `deploy/Dockerfile`: migrations `COPY` is now a genuine wildcard no-op when absent. On the build-info stamp: your suggested `grep -a` approach doesn't actually work — Go embeds the literal `-ldflags` string as build metadata regardless of whether the `-X` target was valid, so grepping the binary can't distinguish correct from wrong (I built this and confirmed it before shipping the fix). What *is* reliably catchable at build time: a revision with no declared `BUILDINFO_PACKAGE` at all, which now fails the build. The "wrong-but-plausible package" case is documented as uncatchable here and left to `check-deployed-revision.sh` at deploy time, which is what it's for. Alpine bumped 3.19→3.22. `scripts/test-dockerfile.sh` builds this against a throwaway project (keel has no `cmd/` of its own) and is wired into a new `deploy-scripts` CI job — all 6 of its checks pass, including the required-to-fail control. |
| 7 | `SingleFlight.Do` now uses `DoChan` + `select` on the caller's own `ctx.Done()`, and runs the shared call on `context.Background()` rather than any one caller's context. Tests cover: a follower with an already-cancelled context returns immediately without blocking; a cancelled caller doesn't poison the result for other waiters; the leader cancelling doesn't cancel `fn`. |
| 8 | `ARCHITECTURE.md` and `README.md` both corrected, and updated to show `media`/`geocode`/`perf`/`scripts`/`deploy` as landed rather than planned. |

### Should-fix (all addressed)

Weak ETag (RFC 9110-correct across content-codings, documented composition order); `ResponseCache` replays status + all headers; `MemoryStore` now bounded by estimated bytes (`MaxBytes`/`MaxEntryBytes`) not entry count; `ETag` capped at `MaxETagBufferBytes` with a streaming fallback past that; `Gzip` has a `MinCompressBytes` threshold and sniffs real content type before deciding (closes the "JPEG gzipped because Content-Type wasn't set yet" case); `NormalizeKey` escapes its field separator (added the exact collision case from the review as a test); `RateLimited` now releases an abandoned reservation on cancellation instead of pushing out the next caller (reproduced your "waited 4s on a 2s interval" symptom as a failing test first); `media/store.go`'s `ErrNotFound` doc corrected; alpine EOL fixed; `LocalStore.Put` temp names are now random per call; `ETag`/`Gzip`/`cacheRecorder` all expose `http.Flusher`.

N1/N4/N5/N8/N13/N20 all now have dedicated tests (verified each goes red under the original mutation before fixing).

### Style

`SingleFlight.Do` returns `(v, shared, err)`; `RateLimitedOptions` removed entirely (unexported `now`/`sleep` fields, constructed via `NewRateLimited` or, in this package's own tests, a direct struct literal); `Provider` un-embedded in `RateLimited`; `Entry.StoredAt` removed (was dead); `TestLocalStore_PathEscapeControl`'s tautological assertion removed — and the `HasPrefix` dead-code note is moot now, since the `os.Root` rewrite replaced that whole mechanism.

### Verification

`go build`/`vet`, `go test -race -count=1 ./...`, `golangci-lint` (now includes `gosec`, caught and fixed a `G301` on a directory-permission call), `govulncheck` (re-pinned `golang.org/x/image` after the rebase reset it to a vulnerable version — 0 reachable now), `gitleaks` (tree + full history), `shellcheck` on all 4 scripts, every script's own `--self-test`, `docker compose config`. CI is green: `test`, `lint`, `vuln`, `secrets`, `deploy-scripts` (new).

Not merging.

ManavA:
## Independent review, round 2 — `909ac6e` — **CHANGES REQUESTED** (one blocker, newly introduced)

Re-reviewed at `909ac6e79377b81303cce7c10fb82b852aa25109`, rebased onto `main` `e58103e`. CI attaches and is green on all five jobs (`test`, `lint`, `vuln`, `secrets`, `deploy-scripts`).

**Seven of the eight blockers are fixed, and fixed properly** — not papered over. The eighth (SingleFlight) fixed the context problem I raised and introduced a worse one in its place. That is the only thing standing between this and an approval.

I also re-ran every one of the six mutations that survived round 1, and **all six now go red**, plus 17 new mutations against the round-2 surface — all red. The suite is genuinely strong now.

---

### The blocker

**A panic inside `SingleFlight.Do`'s `fn` now kills the process.**

Round 1 used `group.Do`, which runs `fn` on the caller's goroutine: a panic propagated to the caller and `httpx/middleware.Recoverer` (or any `defer recover()`) could absorb it. Round 2 switched to `DoChan`, which runs `fn` on a goroutine `x/sync` owns and, on panic, re-panics **from that goroutine**. Nothing in the caller's stack can recover it.

```
=== RUN   TestProbe2_SingleFlightPanic          # caller has defer recover()
panic: boom
    golang.org/x/sync/singleflight.(*Group).doCall.func1.gowrap2()
        .../singleflight/singleflight.go:167
    created by golang.org/x/sync/singleflight.(*Group).doCall.func1 in goroutine 19
```

The caller's `recover()` never ran; the test binary died. For a package whose stated purpose is HTTP middleware, this turns "one handler has a nil-pointer bug" into "the service restarts". It is strictly worse than what it replaced, and it is a regression this round introduced — everything else about the `DoChan` rewrite is right, and I verified the parts you claimed:

```
leader:    v=<nil> err=context canceled
follower1: v=REAL RESULT (fn ctx err=<nil>) err=<nil>      # leader's cancellation does not reach followers
follower2: v=REAL RESULT (fn ctx err=<nil>) err=<nil>      # fn ran on a live context
follower with an already-cancelled ctx returned after 1.186µs, err=context canceled
```

The fix is local — recover inside the `DoChan` closure and turn the panic into a value the caller can act on, rather than letting `x/sync` re-panic off-goroutine:

```go
ch := s.group.DoChan(key, func() (v any, err error) {
    defer func() {
        if p := recover(); p != nil {
            err = fmt.Errorf("perf: singleflight call for %q panicked: %v", key, p)
        }
    }()
    return fn(context.Background())
})
```

(Re-panicking in the caller's goroutine after the receive is the other defensible choice; either is fine, but the current behaviour should not ship, and whichever you pick wants a test — the existing suite does not cover a panicking `fn` at all.)

---

### Blockers 1–8, re-checked

| # | Status | Evidence |
|---|---|---|
| 1 | **Fixed** | alice/bob no longer collide via `Authorization` *or* `Cookie`; an anonymously-warmed entry is not served to an authenticated request (credentialed requests bypass the cache entirely, in both directions). |
| 2 | **Fixed, both orderings.** | `ResponseCache(Gzip(h))` *and* `Gzip(ResponseCache(h))`: gzip client → 75 bytes `CE=gzip`; identity client → the full 10012-byte identity body, `CE=""`; second gzip client → `HIT` with `CE=gzip`. No unlabelled gzip anywhere. |
| 3 | **Fixed, and robustly.** | A symlink planted *after* `os.OpenRoot` is still refused for `Get` (`path escapes from parent`), `Put`, and over HTTP (404) — as is a direct file symlink. `..`, `%2e%2e` and `//etc/passwd` all 404. No directory listing at `/media/dir/`, `/media/dir`, `/media/` or `/media`. `nosniff` on every response, `Content-Type` from the sidecar (`image/svg+xml` served as declared, not guessed). The sidecar and temp files are unreachable; non-GET/HEAD is 405. |
| 4 | **Fixed.** | My exact repro is rejected: `12000x12000 PNG (168667 bytes) -> media: source exceeds max pixels: 144000000 pixels, limit 60000000`. I also probed the animated-GIF angle the coordinator raised: a 300-frame 3000×3000 GIF passes the config gate (one frame = 9 MP), but `imaging.Decode` materialises only the first frame, so decoded memory stays one frame. `MaxPixels` bounds what is actually allocated. No gap. |
| 5 | **Fixed.** | `sanitizeTransportErr` unwraps `*url.Error`; a real dial failure with a distinctive token no longer renders the URL. |
| 6 | **Mostly fixed** — see below. | The wildcard `COPY` is a genuine no-op (verified absent and present); alpine 3.22; `scripts/test-dockerfile.sh` passes 6/6 against real Docker here, including the required-to-fail control and a real `docker run` asserting the stamp reached the container. |
| 7 | **Context problem fixed, panic problem introduced.** | Above. |
| 8 | **Fixed.** | `ARCHITECTURE.md:33` now reads `perf/ response caching, ETag, gzip, and singleflight middleware`, and `media`/`geocode`/`perf`/`scripts`/`deploy` moved into the landed block. README agrees. |

### On the Dockerfile argument — you are right, and then you go one step too far

**You are right about `grep -a`, and I was wrong.** I verified it: the revision string appears in *both* binaries, because `go version -m` metadata carries the literal `-ldflags` string regardless of whether the `-X` target resolved.

```
grep -ac deadbeefcafe /tmp/ok /tmp/bad   ->   /tmp/bad:2   /tmp/ok:3
go version -m /tmp/bad | grep ldflags
        build -ldflags="-w -s -X ldprobe/internal/nonexistent.Revision=deadbeefcafe"
```

Good catch, and `test-dockerfile.sh:122` running the container and asserting the output is the right instrument — that is exactly the check the coordinator asked for, and it is in CI.

**But `Dockerfile:18-27`'s stronger claim — that a wrong `BUILDINFO_PACKAGE` "cannot be [caught] from inside this Dockerfile" — is not true.** Three build-stage checks catch it, in ascending subtlety:

```
go list "$BUILDINFO_PACKAGE"                                  # nonexistent package     -> exit 1
go doc  "${BUILDINFO_PACKAGE}.Revision"                       # no settable Revision    -> exit 1
go list -deps "$CMD_PATH" | grep -qx "$BUILDINFO_PACKAGE"     # real, but not linked in -> the subtlest case
```

I ran all three. The third is the one that matters most: a package that exists *and* has `Revision` but is not in this binary's dependency graph is also a silent `-X` no-op, and only that check finds it. This is a should-fix rather than a blocker — `check-deployed-revision.sh` is a real backstop — but asserting an impossibility that a one-line `go list` disproves is the "comment that outlived its fact" shape, in the one file whose whole job is making the revision stamp trustworthy. Please either add the guard or soften the comment to "grepping the built binary cannot catch this".

### Remaining should-fixes (none blocking)

- **`Cache-Control: no-cache` and `max-age=0` are still cached for the full TTL.** `cacheableResponse` refuses `private` and `no-store` but not these; both mean "do not reuse without revalidation", and the cache reuses for 60s.

  ```
  Cache-Control: no-cache      handlerCalls=1 cached=true    <<< expected refusal
  Cache-Control: max-age=0     handlerCalls=1 cached=true    <<< expected refusal
  Cache-Control: private       handlerCalls=2 cached=false   ok
  ```
- **A response declaring `Vary: <header>` that `DefaultKey` does not fold in is still stored and replayed across that header's values.** `Vary: *` is correctly refused; `Vary: Accept-Language` is cached under a key that ignores `Accept-Language`. Since you already read `Vary`, refusing any `Vary` naming a header the key function doesn't cover would close this the same way `Vary: *` was closed. (`Vary: Accept-Encoding, Cookie` is safe as it stands — the encoding is in the key and cookied requests never reach the cache.)
- **`LocalStore.Put`'s sidecar-failure rollback deletes the object** (`local_store.go:119,124`). On an *overwrite*, that destroys the previous good object too, which contradicts the method's own "never corrupts whatever was there before". There is also a window between the body rename and the sidecar rename where `Handler` serves the object with no recorded content type and falls back to sniffing.
- **Minor:** `Put` through a symlinked directory refuses with `mkdirat link: file exists` — right outcome, misleading reason.
- **Minor:** `COPY --from=builder /src/migrations* /app/migrations` also matches `migrations.md`, `migrations-old/`, etc., merging them into `/app/migrations`.

### Everything else from round 1, confirmed fixed

Weak `W/"…"` ETag (correct across content-codings); status and all headers replayed on a hit (201→201, 204→204, `Cache-Control` preserved); `MemoryStore` bounded by estimated bytes with a per-entry cap; `MaxETagBufferBytes` with a streaming fallback; `MinCompressBytes = 1024` (7-byte body no longer grows to 31) and the real content type sniffed before deciding (a JPEG written with no `Content-Type` is no longer gzipped); `NormalizeKey` escapes its separator (`"a\|b|c||"` ≠ `"a|b\|c||"`); `RateLimited` releases an abandoned reservation (2.00s, was 4.00s); `store.go`'s `ErrNotFound` doc corrected; `RateLimitedOptions` gone; `Entry.StoredAt` gone; `Do` returns `(v, shared, err)`; `http.Flusher` visible behind all three middlewares.

### Mutation results

**Round-1 survivors, all now caught:** N1 remove the `LimitReader`, N4 fixed temp name, N5 re-`Put` recorded variants, N8 swallow tier errors, N13 drop stored headers on a hit, N20 keep the stale `Content-Length` — **6/6 RED**.

**New mutations against the round-2 surface, all caught:** stop refusing credentialed requests; stop honouring `private`/`no-store`; stop refusing `Set-Cookie`; drop the encoding from `DefaultKey`; disable the pixel cap; drop `nosniff`; stop refusing directories; serve the sidecar and temp files; stop sanitizing transport errors; run shared work on the leader's ctx; stop honouring the caller's ctx; issue a strong ETag instead of weak — **12/12 RED**.

The one hole left is the one the blocker names: nothing tests a panicking `fn`.

### Verification run here

`go build`/`vet`/`gofmt -l`, `go test -race -count=1 ./...` (whole module, incl. the Docker-backed `pg/testdb`), `golangci-lint v2.6.0` — 0 issues, `govulncheck` — 0 called, `gitleaks` tree and full history — clean, no GoReal or project identifiers anywhere in the tree, `shellcheck` on all four scripts, all three script `--self-test`s, `scripts/test-dockerfile.sh` 6/6 against real Docker, `go 1.25.0` kept, `golang.org/x/image` back at `v0.44.0`.

Fix the panic path and I will approve.

ManavA:
## Round 3 fixes

Tip: a24d9a1. CI green (test, lint, vuln, secrets, deploy-scripts).

**Blocker — SingleFlight panic crash.** Confirmed the mechanism first: built a throwaway program showing DoChan's fn panicking crashes the process from an uncatchable goroutine (golang.org/x/sync/singleflight spawns `go panic(e)` when more than one caller is waiting, specifically so a caller's own recover cannot stop it), and separately confirmed the fix removes the crash before shipping it. Do now recovers inside the closure passed to DoChan, before the panic ever reaches singleflight's own handling, and returns it as `*ErrPanicked`. Tests: a lone panicking call returns the error without crashing; a follower waiting on a panicking leader gets the same error, not a hang or a second panic (both tests only pass because the process didn't crash while running them).

**Should-fixes, all four:**
- `Dockerfile`: the earlier "cannot be caught from inside this Dockerfile" claim was wrong — verified `go list -deps "${CMD_PATH}" | grep -qx "$PACKAGE"` catches both a nonexistent package and a real one that exists but isn't imported by this build (built and checked both before adding it). Added it, corrected the comment. One residual gap I verified separately and documented honestly rather than overclaim again: a real, reachable package with no variable actually named `Revision` still isn't caught — the linker no-ops that too, silently, which I confirmed by building it. `test-dockerfile.sh` now has 9 checks covering all of this, including that specific documented limitation.
- Narrowed the migrations glob: staged `migrations/` into an always-present builder-stage directory and made the runtime `COPY` unconditional (no glob at all), so nothing named `migrations.md` or `migrations-old/` can be swept in. Added a test with exactly that file present.
- `ResponseCache` now treats `Cache-Control: no-cache` and `max-age=0` as refusing storage (no-cache needs revalidation this cache doesn't implement; max-age=0 is stale on arrival), and generalized the `Vary` check so any header name other than `Accept-Encoding` refuses storage — closes a cross-value replay a `Vary: X-Tenant`-style response would otherwise cause. Reproduced the contamination as a failing test before the fix (bob got alice's response), confirmed it's fixed after.
- `LocalStore.Put`: neither the body nor its content-type temp file is renamed into place until both writes have succeeded, so a failure partway through an overwrite leaves the previous object untouched instead of deleting it. Added a `rootFS` interface seam (same-package only) so a test can inject a failure at the exact write that used to destroy a good object on overwrite — reproduced the original bug against the old ordering first, confirmed the fix against the new one.

Not merging.

ManavA:
## Independent review, round 3 — `a24d9a1` — **APPROVED**

Scoped re-check of the SingleFlight panic blocker and the four should-fixes, at `a24d9a1c0f51c11317b8623c4874ccf5a7580636` (rebased onto `main` `d2d293e`). CI green on all five jobs.

**All five are fixed, and every claim in the response holds up under independent probing.** The Dockerfile answer in particular is now the honest version of the argument: the check that works is implemented, and the one residual is demonstrated rather than asserted.

---

### The blocker — fixed

Recovering inside the `DoChan` closure is the right place, and it holds under the case that worried me most. Eight callers on one key, every one of them panicking:

```
8 callers: 8 got *ErrPanicked, 8 marked shared; sample err=singleflight: panic: boom from 7
a later Do on the same key: v=recovered err=<nil>
follower whose own fn would panic but which joined a healthy leader got: leader value
```

No crash, every waiter gets a typed error with a stack, the key is not left poisoned, and a follower that joins a healthy leader correctly never runs its own `fn` at all. `ErrPanicked` carrying `Value` and `Stack` is better than what I asked for — a caller can log the original stack instead of a flattened string.

Mutation: removing the `defer recover()` → **RED**.

### The four should-fixes — fixed

**`Cache-Control`.** Refused across every spelling I tried, including the `must-revalidate` combination in both orders:

```
""                            cached=true    ok (control)
"public, max-age=60"          cached=true    ok (control)
"no-cache"                    cached=false
"no-cache, must-revalidate"   cached=false
"must-revalidate, no-cache"   cached=false
"No-Cache"                    cached=false
"  no-cache  "                cached=false
"max-age=0"                   cached=false
"max-age=0, must-revalidate"  cached=false
"public, max-age=0"           cached=false
"max-age=00"                  cached=false
"max-age=-1"                  cached=false
"private" / "no-store"        cached=false
```

Mutations: drop the `no-cache` case → **RED**; drop the `max-age<=0` case → **RED**.

**`Vary`.** Exactly as claimed — `Accept-Encoding` only, case- and whitespace-insensitive; `*`, `Accept-Language`, `Cookie`, `Origin`, and every combination refused. Mutation: revert to refusing only `Vary: *` → **RED**.

**`LocalStore.Put`.** Neither temp renames until both writes succeed. Mutation: restore the old write-rename-write-rename ordering → **RED**. The `rootFS` seam is a good call — that failure genuinely is not reachable through a real filesystem.

**Dockerfile.** `go list -deps` is the right gate, and I verified both cases it catches independently of your self-test, by building them:

```
nonexistent package      -> go list -deps grep rc=1  (build fails)
real but unimported pkg  -> go list -deps grep rc=1  (build fails)
```

And I constructed the residual case you document — a package that is real, *is* reachable from `CMD_PATH`, and simply has no `Revision` variable — to check it is genuinely residual rather than conveniently described that way:

```
go list -deps ./cmd/api | grep -qx novarprobe/internal/novar   -> PASSES the gate
go build -ldflags="-X novarprobe/internal/novar.Revision=deadbeefcafe"  -> rc=0
./novar                                                         -> revision: unset
go doc novarprobe/internal/novar.Revision                       -> rc=1   (would catch it)
go doc novarprobe/internal/buildinfo.Revision                   -> rc=0
```

The gate really does let it through, `go doc` really would catch it, and the comment says both of those things and explains why you stopped where you did. That is the right shape: a limitation with its boundary measured, not an impossibility claim. My round-2 objection is fully answered.

`scripts/test-dockerfile.sh` passes 9/9 against real Docker here, including both required-to-fail controls and the `migrations.md` non-match. Staging into `/out/migrations` rather than tightening the glob is the better fix — it removes the failure mode instead of narrowing it.

### Residuals — none blocking, none requiring another round

1. **No test injects a failure at the *first* write.** Both shipped cases use `writesRemaining: 1`. I wrote the `writesRemaining: 0` case and the behaviour is correct — fresh key leaves nothing, overwrite leaves `"original"` / `"text/original"` intact with no temp litter — but nothing in the suite holds that.
2. **The `rootFS` seam declares `Rename` and nothing injects a rename failure.** One mutation survived because of it: re-adding `_ = s.fs.Remove(cleaned)` to the second-rename-failure path — the destructive behaviour you just removed — goes **GREEN**. The shipped code is right; there is no test standing between it and a regression. A `failRename` alongside `failAfterNWrites` closes both (1) and (2); I ran one, and it confirms the documented split: first rename fails → original fully intact; second fails → body updated, content type lags, which is exactly what `Put`'s comment promises.
3. **A failed second rename leaves `ctTmp` behind.** `Handler` refuses `tmpInfix` names so it is litter, not exposure, but it is the one path that does not clean up after itself and the comment does not mention it.
4. **`s-maxage=0` is not refused**, and `ResponseCache` is a shared cache, so that directive is aimed squarely at it. `max-age = 0` with internal whitespace also slips through. Both are narrower than what you fixed; worth a line either way.

### Verification run here

`go build`/`vet`/`gofmt -l`; `go test -race -count=1 ./...` across the whole module including the Docker-backed `pg/testdb`; `golangci-lint v2.6.0` — 0 issues; `govulncheck` — 0 affected; `gitleaks` tree and full history — clean; `shellcheck` on all four scripts; all three script `--self-test`s; `scripts/test-dockerfile.sh` 9/9 against real Docker; `go 1.25.0` kept; `golang.org/x/image` at `v0.44.0`. Sweep for GoReal or project identifiers and for Claude/AI/process references across the tree: **nothing**.

Two things outside this PR, for whoever owns them: a 17.6 MB unstripped ELF binary named `minimal` sits at the repo root, added by `main`'s commit `94f7f45`, not by anything here — `.gitignore` covers `/bin/` but not a bare binary at the root. And the `Co-Authored-By` / `Claude-Session` commit trailers on this branch match `main`'s own convention (8 such lines across its last 6 commits), so that is a repo-wide policy question rather than something this PR introduced.

Good work across three rounds. Approved.

PR #7, opened 2026-09-18
Add retry, idempotency and outbox packages

ManavA:
## Summary

Three new packages, plus a small use of the new one:

- `retry`: `Do(ctx, fn, Options)` with exponential backoff and full jitter (the AWS Architecture Blog's "Full Jitter" algorithm, cited by name in the package doc), a `Retryable` predicate, and immediate stop on context cancellation. Layer 2, no keel imports.
- `idempotency`: HTTP middleware reading `Idempotency-Key`, with a `Store` interface, an in-memory default, and a Postgres-backed implementation in `idempotency/pg` (migration included). Same key + same body replays the stored response with `Idempotency-Replayed: true`; same key + different body gets a 409 that never echoes either body; a request racing a same-key request already in progress waits for it (bounded by `Options.Wait`), rather than getting an immediate conflict — the package doc says why.
- `outbox`: `Enqueue(ctx, tx, Event)` writes an event row inside the caller's own transaction, and `Relay` polls unpublished rows and publishes them through an `events.Publisher` (migration included). Delivery is at-least-once, matching `events/pubsub`'s own documented stance on Pub/Sub; a row's id is the dedupe key consumers use. A row that keeps failing does not block the rows behind it.
- `mail`: `PostmarkSender.Send` now retries a request the HTTP client itself failed to complete (dropped connection, unparseable body), using the new `retry` package. It does not retry a response Postmark returned successfully with a nonzero `ErrorCode` — those are classified as before, and retrying an error like an unknown template alias would just fail the same way every time.

## Mutation checks

Each mutation was introduced, confirmed to fail the relevant test(s), then reverted; the suite is green again at the tip of this branch.

| Package | Mutation | Test(s) that caught it |
|---|---|---|
| retry | Off-by-one in the final attempt's `Attempts` count | `TestDo_ExhaustsMaxAttempts`, `TestDo_NonRetryableErrorStopsImmediately` |
| retry | Doubled `FullJitter`'s starting window | `TestFullJitter_Distribution` |
| idempotency | Skipped the TTL expiry check in `MemoryStore.Claim` | `TestMiddleware_ExpiryReRunsHandler`, `TestMemoryStore_ExpiredRecordIsReclaimed` |
| idempotency | Disabled the request-hash conflict check in the middleware | `TestMiddleware_ConflictOnSameKeyDifferentBody` |
| idempotency/pg | Disabled the in-progress check in `Store.Claim` | `TestStore_ClaimTwiceReturnsInProgress` |
| outbox | Disabled `Enqueue`'s empty-topic check | `TestEnqueue_EmptyTopicErrors` |
| outbox | Removed the `continue` that skips past a failed row | `TestRelay_PoisonedRowDoesNotBlockOthers` |
| mail | Dropped `postmarkTransportRetry.MaxAttempts` to 1 | `TestPostmarkSender_Send_RetriesTransportFailures` |

## Left out, and why

- **Per-aggregate ordered delivery in `outbox`** — out of scope per the package doc: `Event` carries no aggregate key, and adding one plus in-flight tracking per partition is real design work, not a small addition.
- **`geocode/mapbox.go`** — does not exist in this repository; there is no `geocode` package in keel to refactor. (`mail/postmark.go` was refactored; see above.)
- **Multi-instance coordination in `outbox.Relay`** — the fetch query has no `FOR UPDATE SKIP LOCKED`; two Relay instances against the same database can both pick up the same unpublished row. Documented as a limitation rather than solved, since consumers already have to dedupe under the at-least-once contract regardless.
- **Dead-lettering in `outbox.Relay`** — a permanently failing row is retried forever, on every tick; there's no cap on `attempts` that stops trying or moves it aside.

## Test plan

- [x] `go build ./...`
- [x] `go test -race -count=1 ./...` (397 subtests pass, including the 3 new packages' 38 and mail's 2 new ones)
- [x] `KEEL_REQUIRE_DB=1 go test -count=1 ./pg/... ./examples/...`
- [x] `make migrations-check` (both new migration directories are discovered and replay clean)
- [x] `golangci-lint run ./...` — 0 issues
- [x] `govulncheck ./...` — no vulnerabilities in code
- [x] `gofmt -l .` — clean
- [x] `gitleaks detect` — no leaks


ManavA:
## Independent review — CHANGES REQUESTED

Reviewed at `66e8488`, rebased onto current `main` (`d007ef1`) in a scratch clone. Everything below was run against the rebased tree.

### Rebase

Four of the five commits replay cleanly. Only the docs commit conflicts, and only additively:

| File | Conflict | Resolution |
|---|---|---|
| `ARCHITECTURE.md` | layout block — `main` gained `media/ geocode/ perf/` where the branch inserts `retry/ idempotency/ outbox/` | union; the three new entries go before `scripts/`/`deploy/` |
| `ARCHITECTURE.md` | import layers — both sides edit levels 2 and 3 | genuine merge: level 2 becomes `httpx, pg, events, retry`; level 3 becomes `…, media, geocode, perf, idempotency, outbox` |
| `README.md` | package table — `main` removed the `scripts`/`deploy` rows in `6624990` | keep only the three new rows |

No code conflicts, and the layering the branch claims still holds after the merge: `outbox` (level 3) → `retry` (level 2) is legal, `retry` imports nothing.

### Verification

| Check | Result |
|---|---|
| `go build ./...`, `go vet ./...` | clean |
| `go test -race -count=1 ./...` with `KEEL_REQUIRE_DB=1` | all 25 packages pass |
| `make migrations-check` | both new directories auto-discovered — `[examples/minimal/migrations idempotency/pg/migrations outbox/pg/migrations]` — and all three replay clean |
| `golangci-lint run` v2.6.0, **cache cleaned**, cross-checked against the official docker image and a deliberate canary file | 0 issues |
| `govulncheck ./...` | no vulnerabilities in code |
| `gofmt -l .`, `gitleaks detect` | clean |
| authorship / no process references | all five commits `Manav Agarwal <manav@berkeley.edu>`, no trailers; the diff has no forbidden mentions |
| files touched | only the three new package directories plus `mail/postmark.go` and the two doc tables — nothing restructured |

The one lint finding the branch would have added on its own (`goconst` on a third `"msg-1"` in `mail/postmark_test.go`) is absorbed by `d007ef1`'s test exclusion.

### Mutation results

The eight mutations in the description all reproduce as claimed:

| # | Mutation | Claimed catcher | Result |
|---|---|---|---|
| 1 | retry: off-by-one in the final `Attempts` | `TestDo_ExhaustsMaxAttempts`, `…NonRetryableErrorStopsImmediately` | caught, both |
| 2 | retry: doubled `FullJitter`'s starting window | `TestFullJitter_Distribution` | caught |
| 3 | idempotency: skip the TTL check in `MemoryStore.Claim` | `TestMiddleware_ExpiryReRunsHandler`, `TestMemoryStore_ExpiredRecordIsReclaimed` | caught, both |
| 4 | idempotency: disable the request-hash conflict check | `TestMiddleware_ConflictOnSameKeyDifferentBody` | caught |
| 5 | idempotency/pg: disable the in-progress check | `TestStore_ClaimTwiceReturnsInProgress` | caught, plus `TestStore_ConcurrentClaimOnlyOneWins` |
| 6 | outbox: disable the empty-topic check | `TestEnqueue_EmptyTopicErrors` | caught |
| 7 | outbox: remove the `continue` past a failed row | `TestRelay_PoisonedRowDoesNotBlockOthers` | caught |
| 8 | mail: `MaxAttempts` 3 → 1 | `TestPostmarkSender_Send_RetriesTransportFailures` | caught, plus the exhaustion test |

Twenty-eight more of my own. The jitter algorithm is genuinely full jitter and the distribution test genuinely tests a distribution — it kills a no-jitter mutant, an **equal**-jitter mutant (`window/2 + rand(window/2)`, which still has the right mean), a fixed-offset mutant, a linear-backoff mutant and a shifted-exponent mutant. These survived:

| # | Mutation | Survived because |
|---|---|---|
| S1 | `mail`: retry a parsed nonzero `ErrorCode` too | **nothing pins the headline mail claim.** The `ErrorCode` tests assert the returned error, never the call count, so "a permanent failure is sent once" is unverified |
| S2 | `mail`: pass `context.Background()` to `retry.Do` instead of `ctx` | the ctx plumbing this PR added to `Send` is untested |
| S3 | `idempotency`: drop `r.Body = io.NopCloser(...)` | **no test has a handler that reads the request body**, so the middleware's most invasive side effect — consuming `r.Body` — is unpinned |
| S4 | `idempotency`: replay without the stored headers | no test asserts a replayed header |
| S5 | `idempotency`: let a second `WriteHeader` overwrite the first | `responseCapture`'s once-only guard is unpinned |
| S6 | `idempotency/pg`: `Release` deletes completed rows too | the `and done = false` guard is unpinned |
| S7 | `outbox`: `Run` never returns on cancellation | **`Relay.Run` has no test at all** — ticker, shutdown and error logging are uncovered |
| S8 | `outbox`: ignore `BatchSize` | the option is unpinned |
| S9 | `outbox`: drop `order by created_at` | creation-order delivery, which the package doc asserts, is unpinned |
| S10 | `retry`: replace the cancellable timer with `time.Sleep` | see blocker 3 — the test catches this only ~11 times in 12 |

---

## Blockers

### 1. An abandoned claim wedges an idempotency key permanently, and a client hanging up is enough to cause it

`idempotency/pg/store.go`'s claim statement reclaims only `done = true AND expires_at < now()`. A row with `done = false` is never reclaimable, at any TTL, by any instance. `MemoryStore.Claim` has the same shape: it returns `ErrInProgress` for an unfinished entry without ever consulting that entry's own `ExpiresAt`, which `Claim` took the trouble to set. There is no lease, no heartbeat and no reaper anywhere in the package or the migration.

```
claim with a 1ms TTL, wait 50ms, claim again
  → idempotency: request already in progress
```

The store doc already concedes this ("the key is stuck in progress until it is reclaimed by an implementation that expires abandoned claims on its own") — but neither shipped implementation is that implementation, so the escape hatch is empty.

What turns this from a crash-only edge case into a routine one is that `Middleware` hands `r.Context()` to both `Complete` and the `Release` that cleans up after it. When a client hangs up — exactly what happens to the slow POSTs people put idempotency keys on — that context is already cancelled, so both writes fail and the key is stranded:

```
Complete on a cancelled request context: idempotency/pg: complete "…": context canceled
Release on the same cancelled context:   idempotency/pg: release "…": context canceled
the key afterwards:                      idempotency: request already in progress
```

End to end, every later request for that key gets a 409 after burning the full `Wait`, forever.

Two changes, both small:

- give the claim a lease the reclaim honours — either a separate `claimed_at`/`lease_until` column, or simply let the claim row's own `expires_at` reclaim it (`where expires_at < now()`) with the claim taken at a lease duration rather than the full response TTL. `MemoryStore` needs the matching check against `e.record.ExpiresAt` on the `!e.done` path.
- use `context.WithoutCancel(r.Context())` for `Complete` and `Release`. The handler has already run; the bookkeeping must not be cancelled by the client that is no longer listening.

A test that claims, abandons, waits past the lease and reclaims would have caught the first. A test that completes on a cancelled context would have caught the second.

### 2. The request hash covers the body only, so one key crosses endpoints

`hashBody(body)` hashes the body and nothing else. Two different endpoints reached with the same key and the same body replay each other:

```
POST /charge  -> 201 "charged"
POST /refund  -> 201 "charged"  Idempotency-Replayed: true
```

The refund never runs and the client is told it succeeded. Keys are client-generated, so a client that scopes a key per logical operation rather than per endpoint — or that reuses one for a retry of a different call in the same saga — hits this. Empty-body POSTs to different routes are the worst case: they all hash identically.

Hash `method + "\n" + path + "\n" + body`, or store method and path on the record and 409 on mismatch the same way the body mismatch does. Either is a few lines; the existing `TestMiddleware_ConflictOnSameKeyDifferentBody` extends to cover it.

### 3. Filling a batch with poison rows starves the outbox completely, and the doc claims the opposite

`outbox/doc.go` says a failing row "is retried on its own while later rows keep flowing past it", and `Tick`'s doc says "one row stuck failing does not block the rows after it". That holds only while the failing rows are fewer than `BatchSize`. `fetch` selects the oldest `BatchSize` unpublished rows every tick with no attempts filter, so once that many rows are permanently bad they occupy the whole batch forever:

```
BatchSize 5, five poisoned rows, one healthy row behind them
after 10 ticks: the healthy row has attempts=0 and is still unpublished
```

Not merely delayed — never even attempted. The default `BatchSize` is 20, so twenty bad payloads (one broker rejecting a topic, one oversized message) silently stop all delivery, and the `attempts` column that would reveal it is written but never read.

The "left out" list is honest that there is no dead-lettering, and I agree that is a design decision rather than a bug. The bug is that the doc promises isolation the implementation does not provide once the batch fills. Either fix the liveness — order by `attempts` then `created_at`, or skip rows over an attempts threshold in `fetch` so they cannot monopolise the batch — or, at minimum, correct both doc claims to say the isolation holds only below `BatchSize` failing rows. I would rather see the ordering change: it is a one-line `order by` and it makes the doc true.

---

## Should fix

### 4. The cancellation test is a coin flip

`TestDo_CancellationStopsBeforeNextAttempt` asserts `elapsed < 2s` against a sleep whose length is `rand.N(5s)` — the very randomness under test. Replacing the cancellable timer with a plain `time.Sleep` still passes whenever the draw comes in under two seconds; across twelve runs of that mutant it survived once.

The bound does not need to be loose. Cancellation happens at 50ms and the real implementation returns immediately, so `assert.Less(t, elapsed, 250*time.Millisecond)` is just as safe for the correct code and leaves the mutant a 5% escape instead of a 10-40% one.

### 5. Extend the transport retry to `geocode/mapbox.go`

The "left out" list says `geocode/mapbox.go` "does not exist in this repository". It does now — `main` gained it in #3 — so that line needs to go either way. Having rebased and looked: there is no ad-hoc retry there to unify, but there is the same gap `mail` just closed, and it is doing visible damage.

`MapboxProvider.Geocode` degrades address → postcode → place, treating a transport failure on one tier as a reason to fall to the next. So a single dropped connection silently returns a *less precise* answer with a nil error:

```
one dropped connection on the address tier
  → Coordinates{Precision: "postcode"}, err = nil
```

The caller cannot tell a blip from an address Mapbox genuinely cannot match. Wrapping `p.httpClient.Do` in `retry.Do` with the same transport-only scope as `postmarkTransportRetry` is about six lines, needs no transient/permanent status classifier (it retries only errors from `Do`, exactly as `mail` does), and the GET carries no body so replay is safe. `geocode` is level 3 and `retry` is level 2, so the import is legal under the layering this PR just edited.

### 6. A Postmark 5xx carrying a parseable body is treated as permanent

`keighl/postmark`'s `doRequest` never looks at the HTTP status — it returns an error only when the transport fails or the body will not unmarshal. So a 503 whose body is valid Postmark JSON parses as an ordinary response with an `ErrorCode`, and the new retry declines it:

```
HTTP 503, body {"ErrorCode":100,"Message":"service unavailable"}
  → calls=1, err=postmark error 100: service unavailable
```

That is the transient failure a retry exists for, and it is the one case the new code does not cover. The doc's reasoning ("retrying an unknown template alias would just fail the same way") is right about *permanent* codes but is being applied to all of them. Either classify a small set of transient codes as retryable, or — cleaner — say so in the doc, since a reader of "retries a request the client itself failed to complete" would not guess that a 503 is excluded.

Also worth two tests regardless (see S1/S2): one that counts calls for a permanent `ErrorCode`, and one that passes a cancelled context. The central mail claim in this PR currently has no test behind it.

### 7. `MemoryStore` never evicts

Nothing removes an entry except `Release` or a repeat claim of the same key. Keys are per-request, so a server on the default store grows without bound:

```
1000 records, every one expired an hour ago
→ 1000 of 1000 still resident
```

At the 24-hour default TTL with response bodies stored whole, this is the in-process default quietly becoming a leak. A size cap with LRU eviction, or a sweep on claim, would do — `perf`'s `MemoryStore` already has the cap-and-evict shape to copy.

### 8. Three contracts the docs should state

- **`Middleware` buffers everything.** `io.ReadAll(r.Body)` has no size limit and nothing in the toolkit imposes one, and the whole response is buffered before anything reaches the client. An 8 MiB body sails straight through. Say that the middleware is for bounded request/response bodies and that a caller should put `http.MaxBytesReader` in front of it.
- **`responseCapture` drops `http.Flusher`.** A handler behind the middleware gets `w.(http.Flusher)` → false and `http.NewResponseController(w).Flush()` → error, so SSE and chunked-progress handlers stop streaming the moment the middleware is added. Inherent to replay; it just needs saying.
- **`Envelope` base64s the payload on the wire.** A subscriber receives `{"id":"…","payload":"eyJvcmRlciI6MX0="}`. That follows from the struct tags, but `events.Marshal`'s documented promise is that a `[]byte` passes through *unchanged*, and `TestEnqueue_BytePayloadStoredVerbatim` reinforces the expectation for storage. One line on `Envelope` saying the consumer decodes `payload` closes it.

---

## Confirmed as described

- **`retry`** is the AWS Full Jitter algorithm: uniform in `[0, min(cap, base·2^(attempt-1)))`, with the 1-indexing stated in `FullJitter`'s own doc. `MaxAttempts` ≤ 0 and `BaseDelay` ≤ 0 both fall back to the defaults; a `Retryable` that rejects the first error runs exactly one attempt and does not sleep; an already-cancelled context runs nothing and reports `Attempts: 0`; the doubling loop guards `time.Duration` overflow.
- **`idempotency`** concurrent same key, same body: exactly one handler execution, the other waits and replays the identical response with `Idempotency-Replayed: true` — not a 409, as designed. Same key, **different** body concurrently is deterministic, not racy: across 30 races, always exactly one execution and exactly one 409. Headers, status and body all replay exactly. A key reused after expiry re-runs the handler. `-race` clean throughout, including the 8-way concurrent Postgres claim.
- **`outbox`** `Enqueue` inside a rolled-back transaction leaves no row. Two relays against one table **do** double-publish — I reproduced it rather than taking the doc's word, and both deliveries carry the same id. Shutdown mid-tick leaves no torn state: a row whose mark-published loses the race stays unpublished and is redelivered. There is genuinely no attempts cap and no dead-letter.
- **`mail`** a real connection refused (nothing listening) retries three times over ~400ms; a permanent `ErrorCode` on a successful HTTP response is sent exactly once, for 406, 1101 and 300 alike.
- **Migrations** both directories are picked up by the replay gate from #3 with no registration step, and both survive up → up → down → up.

## Notes, not asks

- `retry.Do`'s unreachable tail return carries an error string that can never be observed. Making the loop `for attempt := 1; ; attempt++` — the `attempt == maxAttempts` check inside the body already terminates it — removes both the dead return and the comment explaining why it is dead.
- `Do` defaults `BaseDelay` and then `FullJitter` defaults it again; the mutation that removes the first is behaviourally identical. Harmless, but one of the two is redundant.
- `Retryable` is not consulted on the final attempt. Deliberate, and `TestDo_RetryableSeesEachError` pins it, but it is the kind of contract the CHANGELOG's "differs from the obvious behaviour" list exists for — along with idempotency's wait-instead-of-409 and outbox's at-least-once. #3 did not update the CHANGELOG either, so this is a repo-wide gap rather than this branch's.
- Comment density is in good shape. The long blocks are all package docs in `doc.go`, which is the convention here; the in-body comments explain decisions — why an expired entry falls through, why a nil body becomes `[]byte{}`, why a failed mark means republication — rather than restating the code. Nothing to trim.

ManavA:
## Round 2 — CHANGES REQUESTED

Reviewed at `fa6caee`. All three round-1 blockers are genuinely fixed — I reverted each fix myself and watched the failure come back rather than taking the description's word for it. Two of the fixes, though, introduce new failures that are worse than what they replaced.

### Round-1 blockers: verified fixed

I reverted each change in isolation and reran the suite:

| Reverted change | Result |
|---|---|
| `idempotency/pg`: put `done = true and` back in the reclaim `WHERE` | `TestStore_AbandonedClaimBecomesReclaimableAfterLeaseExpires` fails |
| `MemoryStore`: check in-progress before expiry again | **no test fails** — see gap G1 |
| `Middleware`: `r.Context()` instead of `context.WithoutCancel` | caught |
| hash back to body-only | `TestMiddleware_SameKeySameBodyDifferentRouteIsNotReplayed` and the empty-body variant fail |
| `fetch` back to `order by created_at` | `TestRelay_PoisonedRowsExceedingBatchSizeDoNotStarveAHealthyRow` fails |
| retry cancellation bound back to 2s | caught |
| drop the geocode transport retry | two new subtests fail |
| drop `postmark5xxTransport` | `TestPostmarkSender_Send_RetriesA5xxWithAParseableBody` fails |
| drop `evictExpiredLocked` | **no test fails** — see gap G2 |

And behaviourally, against my round-1 probes:

- `POST /charge` then `POST /refund`, same key and body → the refund is now a 409, and the refund handler is never called. Blocker 2 closed.
- An abandoned claim with a 1ms TTL is reclaimable after 50ms in both stores. Blocker 1 closed.
- An end-to-end client disconnect through the middleware no longer strands the key: a later request replays normally and the handler ran exactly once. `context.WithoutCancel` works.
- A healthy row behind 5 poisoned rows at `BatchSize` 5 now publishes. Blocker 3 closed for a static table.
- A 503 with a parseable body is now retried 3×. A permanent `ErrorCode` is still sent exactly once.
- The retry cancellation mutant is now caught **20 times out of 20** (it was 11 of 12 at the old bound).

### CI

Green on GitHub for this tip, not just locally — run [35312380987](https://github.com/ManavA/keel/actions/runs/35312380987), conclusion `success`, all five jobs: `lint` 1m35s, `test` 5m18s, `secrets`, `vuln`, `deploy-scripts`. The diff touches no Go version pin (`go.mod`, `go.sum`, `.github/`, `Makefile` are all untouched between `b8fea37` and `fa6caee`), so the toolchain-skew failure mode from the other branch does not apply here. Locally: `gofmt`, `go vet`, `-race` suite and `make migrations-check` all clean, and `golangci-lint` reports 0 issues from a cleaned cache.

---

## New blockers

### N1. The outbox ordering fix trades one starvation mode for a worse one

`order by attempts, created_at` fixes the poisoned-batch case by making a retried row sort behind untried rows. That is exactly the problem: **a row that fails even once sorts behind every row that has not been tried yet, including rows enqueued long after it.** Under a steady stream of arrivals it is never selected again.

One row, one transient failure, then a broker that would accept it fine:

```
tick 1: the row fails once            → attempts = 1
then 40 ticks, 5 new events per tick, BatchSize 5
after 40 ticks: the once-failed row is published = false
                unpublished backlog  = 1   (it is the only row left)
```

Every tick the batch is filled by five `attempts = 0` arrivals and the `attempts = 1` row is row six in the ordering, forever. Nothing about it is poison — one dropped connection to the broker is enough, and the outbox's whole purpose is that an enqueued event is eventually delivered.

This is strictly new: under `order by created_at` a once-failed row kept its place and was retried on the very next tick.

The standard fix also happens to be the one that solves *both* modes: give the table a `next_attempt_at` and back off on failure —

```sql
where published_at is null and (next_attempt_at is null or next_attempt_at <= now())
order by created_at
```

A poisoned row backs off exponentially and stops occupying batches; a once-failed row rejoins creation order as soon as its backoff elapses. It also gives `attempts`, which the table already stores and nothing reads, something to do. If you would rather keep the current shape, `order by least(attempts, 3), created_at` at least caps how far a row can fall behind — but it only bounds the damage rather than removing it.

Relatedly, the package doc's bound ("within `ceil(P/B)` ticks") is derived for a static table. I confirmed it holds with no arrivals — the healthy row published after 3 ticks against a bound of 4 — but it is not a general bound and the doc reads as though it were.

### N2. The lease and the replay TTL are the same value, so a short TTL runs the handler twice

`ttl` now does double duty: the claim's lease and the completed record's replay window. When the lease expires under a still-running handler, a second request reclaims the key and runs the handler again:

```
TTL 30ms, handler takes 120ms, two requests with the same key and body
handler executions for one idempotency key: 2
```

That is the single thing this middleware exists to prevent, and it is reachable from a plausible configuration. `Options.TTL` is documented as "how long a completed response is replayed before the key can be reused" — an operator shortening it to bound table growth has no way to know they are also shortening the mutual-exclusion window to less than their p99 handler latency. The default 24h hides it; a five-minute TTL on a handler that occasionally takes six minutes does not.

**And the steal corrupts the record.** `Complete` is `update … where key = $1` with no fencing, and it writes `status`, `header`, `body` and `expires_at` but **not** `request_hash`. So when the request that lost its claim finishes and completes anyway, it lands on the winner's row:

```
A claims (50ms lease) → lease expires → B reclaims with hash-B → A completes
record now: stored hash = "hash-B", stored body = "request A's response"
```

B's client sends B's request, its hash matches, and it is handed A's response body as a valid replay. Silent cross-request response disclosure, and the 409 that is supposed to catch a hash mismatch cannot fire because the hash on the row is B's.

Both symptoms come from one root cause. The fix is to separate the two durations and fence the write:

- a separate `Options.ClaimLease` (short — handler-timeout scale, defaulting to something like a minute) for reclaiming an abandoned claim, distinct from `TTL` for the replay window. The `Store.Claim` signature already takes `ttl`; it needs the lease alongside it, or a claim-specific value.
- a fencing token on the claim — a `claim_id` written at claim time and required by `Complete` (`where key = $1 and claim_id = $2`) — so a request that lost its claim gets `no claim found` instead of overwriting. `Complete` already returns an error on `RowsAffected() == 0`, so the plumbing is there.

`MemoryStore` needs the matching treatment.

---

## Should fix

### N3. The query string is still not in the hash

The hash now covers `r.Method` and `r.URL.Path`, which closes the cross-route case. It does not cover `r.URL.RawQuery`, so two genuinely different requests that differ only in their query still replay each other:

```
POST /transfer?to=alice&amount=10    -> 201 "handled to=alice&amount=10"
POST /transfer?to=mallory&amount=1000 -> 201 "handled to=alice&amount=10"  Idempotency-Replayed: true
handler saw queries: [to=alice&amount=10]
```

Same key, same body, and the second request never reaches the handler. This is the same class as the blocker that was just fixed, and the fix is one more `h.Write` in `hashRequest`. `r.URL.RawQuery` is the right field — it is stable across a retry of the identical URL, so it introduces no false conflicts.

To answer the specific worry directly: **the method+path change does not cause false 409s.** Five successive retries of the identical request all replay with `Idempotency-Replayed: true` and exactly one handler execution.

### N4. `evictExpiredLocked` walks the whole map on every `Claim`, under the store's only mutex

The eviction fix is correct but O(n) per call, and `Claim` is on every request path:

```
     100 entries:     2.2 µs/claim
  10,000 entries:      87 µs/claim
 200,000 entries:     7.0 ms/claim
```

Seven milliseconds of lock-held work per request, and it degrades as the map grows. With the 24-hour default TTL, 200k resident keys is about 2.3 requests/second sustained — so the in-process default becomes a throughput ceiling at ordinary traffic, which is the opposite of what the fix intended.

A bounded partial sweep fixes it in a few lines: evict at most a fixed number of entries per call, or only sweep when at least a second has passed since the last one. Either keeps the memory bound while making `Claim` O(1) amortised.

### N5. The commit message names the pull request

```
fix(idempotency,outbox,mail,geocode): address code review findings on PR #7
```

The PR number does not belong in the permanent history, and "address code review findings" describes the process rather than the change. Something like `fix(idempotency,outbox): lease abandoned claims and keep poisoned rows from starving the outbox` says what changed.

Also, the description says the branch was squashed to one commit; it is six — the round-2 work sits on top of the original five. Worth reconciling before merge if a single commit is what you want.

---

## Test gaps still open

| # | Mutation that survives | What is unpinned |
|---|---|---|
| G1 | `MemoryStore.Claim`: check in-progress before expiry again | **the MemoryStore half of blocker 1 has no test.** The pg half does. The fix is correct — I verified it behaviourally — but reverting it is silent, and `MemoryStore` is the default |
| G2 | remove `evictExpiredLocked` | the eviction fix has no test |
| G3 | `mail`: pass `context.Background()` to `retry.Do` | the ctx plumbing in `Send` |
| G4 | `idempotency`: replay without the stored headers | header replay fidelity |
| G5 | `idempotency/pg`: `Release` without `and done = false` | the guard against releasing a completed row |
| G6 | `outbox`: ignore `BatchSize` | the option |
| G7 | `outbox`: `order by attempts` without `created_at` | the creation-order tiebreak |

Three of round 1's gaps are now closed and pinned by real tests: the permanent-`ErrorCode` call count, the request-body restoration (`TestMiddleware_WrappedHandlerCanReadRequestBody`), and `Relay.Run` cancellation. Note that the `Run` mutant is caught by hanging until the package timeout rather than by an assertion — it does fail the build, but a `Run` call with its own short deadline in the test would fail in milliseconds instead of ten minutes.

---

## Also confirmed unchanged

Two relays against one table still double-publish, with the same id both times — documented, and correct under the at-least-once contract. `Enqueue` under a rolled-back transaction still leaves no row. Shutdown mid-tick still leaves no torn state. The geocode retry is properly context-bounded: a total outage takes ~670ms on its own and exactly 300ms under a 300ms caller deadline, so the per-tier retry does not escape the caller's timeout.

The three documented constraints (unbounded buffering, `Flusher` dropped, `Envelope` base64) all read well and sit next to the code they describe.

